PR #60, opened 2011-07-12
Added SchemaNameAttribute and related handling.  Added dbType mapping for

ChrisMH:
Added SchemaNameAttribute and related handling.  Added dbType mapping for DevArt's PostgreSQL driver.

I don't know if you're interested in these.  I use PostgreSQL's schemas extensively and having a separate schema attribute made sense for me.

I like the automated way you can have the framework add the SELECT and FROM clauses in some situations.  Are you planning any changes to extend that to more complicated queries such as JOINs?

Chris


pleb:

PR #111, opened 2012-04-20
Add support for datetimeoffset, nullable enums, and non-integer backed enums.

ianobermiller:
If you have a column with type datetimeoffset, it will just be mapped to a string in the T4 template. This fixes it to map to the DateTimeOffset CLR type. Also fixes datetime2 to be mapped to DateTime.

Also add support for nullable enums and non-integer backed enums as your POCO datatypes.


pleb:

PR #116, opened 2012-05-14
Let tt files work with MonoDevelop

lebowitz:
MonoDevelop does not have EnvDTE assembly available, but I removed the .config detection code and the rest works.  The only extra thing is that the top-level tt file now needs both ConnectionString and ConnectionStringName both specified.


pleb:

PR #120, opened 2012-06-05
Added GetFactoryKey to prevent func generation on same DataReader results but different sql strings

firestrand:
...nt case where sql string has a value injected resulting in many sql strings with same DataReader results, which creates a new func for each, eventually causing a failure i.e. someone using String.Format to put an id into the SQL string for each call.


pleb:

PR #121, opened 2012-06-13
match only latest order by in query to get count query

vlko:
this patch will allow you to use order by in subqueries (we have one nasty report query containing this)


pleb:

Issue #124, opened 2012-06-25
SkipTake<String>(int, int, PetaPoco.Sql) returning the wrong column


jodydonetti:
Hi, i'm using PetaPoco v4.0.3 and doing this:

``` c#
db.SkipTake<String>(10, 20, PetaPoco.Sql)
```

where the PetaPoco.Sql variable contains the following SQL code, generated using the helper methods:

``` sql
SELECT ColumnName1 FROM TableName ORDER BY ColumnName2
```

I've tracked down the problem to this: once the sql query has been executed and it is creating the instances for each rows, it gets the value from the first column (i assume it does this because of the fact that `<T>` is a primitive type) but, since i'm using SkipTake<T> , it builds the final sql code adding the ROW_NUMBER() stuff and it adds, as a first column, the `peta_rn` column.

The code i'm talking about is this:

``` c#
sqlPage = string.Format("SELECT * FROM (SELECT ROW_NUMBER() OVER ({0}) peta_rn, {1}) peta_paged WHERE peta_rn>@{2} AND peta_rn<=@{3}",
          sqlOrderBy==null ? "ORDER BY (SELECT NULL)" : sqlOrderBy, sqlSelectRemoved, args.Length, args.Length + 1);

```

Now, i've tryied to simply put the `peta_rn` field at the end of the column list, like this:

``` c#
sqlPage = string.Format("SELECT * FROM (SELECT ROW_NUMBER() OVER ({0}) {1} , peta_rn) peta_paged WHERE peta_rn>@{2} AND peta_rn<=@{3}",
          sqlOrderBy==null ? "ORDER BY (SELECT NULL)" : sqlOrderBy, sqlSelectRemoved, args.Length, args.Length + 1);

```

but then the problem moved forward, i suppose in one of these 2 points:
1. when you are IL-emitting the low-level stuff
2. inside the FindSplitPoint(...) method

At feeling, i think it may depends from the fact that normally (deserializing complex types) it removes the first column but, when doing the same for simple types, it doesn't do that.

Am i right? Is it possible to fix that?

Thanks, and btw excellent work so far!

PR #125, opened 2012-06-27
Moving the peta_rn column at the end, to fix the issue #124 v5?

njy:


pleb:
Does need to be fixed in v5?

pleb:
#124

pleb:
I've had a quick look at the code. I'm going through writing integration tests. I'll spend some time with this when I get the paging

PR #130, opened 2012-07-13
T4 code to generate the classes from the Sqlite DB

gkathire:
This is a fix for the T4 code to generate the classes from the Sqlite DB


pleb:

PR #137, opened 2012-10-04
Output parameters bug fixed - (only with named parameters)

JLoustau:
Allows petapoco to manage repeated occurrences of an output parameter in a sql statement
For example the following code was not supported:
private static void TestOutputParamNoSP()
{
            var param = new SqlParameter("@p", SqlDbType.Int);
            var i = 0;
            param.Direction = ParameterDirection.Output;
            PetaPocoTestsDB.GetInstance().Execute(@"
SET @i= @i + 58
set @p = @i + 15
set @p = @p* 2
", new { i, p = param });
            Console.WriteLine("Output param value={0}", param.Value);
        }

### DIFF
--- a/PetaPoco/PetaPoco.cs
+++ b/PetaPoco/PetaPoco.cs
@@ -381,6 +381,12 @@ namespace PetaPoco
 					}
 					return sb.ToString();
 				}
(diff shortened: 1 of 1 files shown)


pleb:

PR #145, opened 2012-12-01
Accept connectionString (not name) and providerName in generated constructor

Finzzownt:
Allows initialization using a connectionString (not a name) and provider. Handy for situations where the connectionString must be constructed at runtime.


pleb:

PR #146, opened 2012-12-02
Generation of MySQL objects scoped to active database

Finzzownt:
When generation objects from a MySQL database, the selection of objects scopes to only objects from the current or specified (in Database.tt) schema name.


pleb:

PR #149, opened 2012-12-26
Updates codebase to support Medium/Partial trust

Shazwazza:
We had read somewhere that PetaPoco was supposed to support partial trust but unfortunately this is not the case. The problem is in only one part of the code in the CreateMultiPocoFactory method. This method creates a DynamicMethod that is not medium trust compatible. There are only 2 overloads that work for creating a DynamicMethod in medium trust and those are the overloads that don't set an 'owner' for the method. By not setting an owner, the method is essentially a static anonymously hosted method. It turns out that this DynamicMethod in particular is actually not required and therefore the IL Emit code is not required either. I did have it working with IL Emit codes in medium trust before I realized this so I still have that code saved but it is much nicer (and probably better performance wise)  to not have to use IL Emit.

An overview of the changes made:

- changes CreateMultiPocoFactory, this is the problem with medium trust... you cannot declare a DynamicMethod with an 'owner' in medium trust. Also, there there is no need for any of the IL.Emit logic in this method and therefore no need for the DynamicMethod at all in this method. It has been changed so that the MultiPocoFactory class does the heavy lifting. Without a DynamicMethod or IL Emit, this should work a teeny bit faster and is much easier to debug. Before I realized that the IL Emit was not necessary, i did make it work in medium trust with IL Emit so have saved that code locally in case its required.
- changes some 'object' references to 'Delegate' (easier to understand)
- changes MultiPocoFactory class to perform the operation of calling the callback and calling each mapper delegate

### DIFF
--- a/PetaPoco/PetaPoco.cs
+++ b/PetaPoco/PetaPoco.cs
@@ -841,7 +841,7 @@ namespace PetaPoco
 		public IEnumerable<T1> Query<T1, T2, T3, T4>(Sql sql) { return Query<T1>(new Type[] { typeof(T1), typeof(T2), typeof(T3), typeof(T4) }, null, sql.SQL, sql.Arguments); }
 
 		// Automagically guess the property relationships between various POCOs and create a delegate that will set them up
-		object GetAutoMapper(Type[] types)
+		Delegate GetAutoMapper(Type[] types)
 		{
 			// Build a key
 			var kb = new StringBuilder();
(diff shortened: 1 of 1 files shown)


pleb:

Shazwazza:
The thing about this change though is that in theory it is a performance optimization. This change essentially means there is less code generation required because the actual change means that a statically declared method is used instead of the same method being generated each time. 

We've been using this change in Umbraco core for years now. I've not actually run the benchmarks on this change as compared to the original code, so I'm not sure how much perf optimization there is but there's definitely less codegen going on.

pleb:
I've reopened the PR. Thanks for the explanation.

PR #159, opened 2013-05-17
Default values from SQL columns

stuartpittaway:
Hello, I've modified the T4 code to automatically generate constructor classes and init the field values based on the database columns.

The problem I was getting was when working with existing tables (with many hundreds of columns) but only wanting to insert a few column values.

This change prevents me having to manually set each field in my code.

Regards

Stuart.


pleb:

PR #161, opened 2013-06-06
Add decimal primary key

rwbaskette:
I have several databases with the decimal type as it's primary key. It originally had to do due with SQL Server 6.5's lack of a bigint data type and has been long carved in stone.

I end up patching it every time I look for an update and and am submitting it in the hope that it will help someone else too.

Thanks!


pleb:

PR #164, opened 2013-09-26
Fix duplicate escape sql identifier V5

NetRube:
fix duplicate escape sql identifier


pleb:
First, thank you for contributing. PP relies on people, such as yourself.

I'm not sure how you're getting the double escape issue. I'm currently adding integration tests for PP. If I come across this issue I'll fix. Else, please PR an integration test to show the problem.

PR #165, opened 2013-10-02
Only mark column as modified if it is a different value.

eldiosyeldiablo:
Before setting the property with the value check to see if it is really
an updated value.
When it is take the value and then call MarkColumnModified.


pleb:

PR #166, opened 2013-10-04
ODBC support

simonhaines:
Parameter substitution differs by using '?' instead of '@0', '@1' etc.
Parameter order is important.


matthewsheeran:
Yeah I have hit this issue too in my Odbc Database Provider support I added to my PetaPoco: any hints on where to look and which members to modify for this as there is an awful lot of parameter processing code and not just @0.. but also @names.. which presumably get translated to the former so the Sql Client can handle the substitution. At 7688 LoC in the PetaPoco.cs I would call it small.
There are hooks for pre and post processing Database requests so that might be the way to start and translate the '@n's to '?'s perhaps...

matthewsheeran:
Weer I to code my own micro Orm from scratch I would have perfomed this substitution in the code  outside of the Stack and avoided the differences across Database Providers that way.

BTW: I would be happy to help with the ODBC Support!?

PR #167, opened 2013-10-04
ODBC support

simonhaines:
Parameter substitution differs by using '?' instead of '@0', '@1' etc.
Parameter order is important.


pleb:

PR #172, opened 2013-11-09
Adds support for specifying isolation level both globally and at the transaction level. v5

antinescience:
So in [this StackOverflow question](http://stackoverflow.com/questions/19862748/petapoco-setting-transaction-isolation-level), I asked about the possibility of specifying an isolation level for transactions while using PetaPoco. Since it looked like - at least on initial glance - there wasn't any way to do this without actually prepending `"SET TRANSACTION LEVEL [LEVEL]"` before each query, I added in a way to specify isolation level.

At the transaction level:

```csharp
var newObject = new NewObject();
var newObjectId = 1;
using (var scope = db.GetTransaction(IsolationLevel.ReadUncommitted))
{
    newObject = db.SingleOrDefault<NewObject>("SELECT * FROM tblNewObject WHERE Id = @0", newObjectId);
    scope.Complete();
}
```

...and at the database level (specifically, in this case, using Autofac as the IoC):

```csharp
builder.Register(a => new PetaPoco.Database("connectionStringName") { DefaultIsolationLevel = IsolationLevel.ReadUncommitted }).As<PetaPoco.Database>().InstancePerLifetimeScope();
```

I would like to expand out to being able to provide an isolation for the LINQ-esque methods without having to explicitly manage transactions, but without further examination of the source, I can't determine whether or not that would have adverse effects. For now, I think this is a good way to be able to provide some level of transaction isolation control.

### DIFF
--- a/PetaPoco/Core/Transaction.cs
+++ b/PetaPoco/Core/Transaction.cs
@@ -2,6 +2,7 @@
 // Copyright © 2011-2012 Topten Software.  All Rights Reserved.
  
 using System;
+using System.Data;
 
 namespace PetaPoco
 {
@@ -18,8 +19,14 @@ namespace PetaPoco
 		public Transaction(Database db)
 		{
 			_db = db;
-			_db.BeginTransaction();
+			_db.BeginTransaction(db.DefaultIsolationLevel);
 		}
+
+        public Transaction(Database db, IsolationLevel isolationLevel)
+        {
+            _db = db;
+            _db.BeginTransaction(isolationLevel);
+        }
 
 		public void Complete()
 		{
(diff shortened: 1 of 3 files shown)


pleb:
Thanks for contributing! It'll be a few days before I can properly review this

antinescience:
Sure. It's been awhile since I looked at it but let me know if there are questions.

pleb:
@CollaboratingPlatypus/petapoco-code-reviewers

pleb:
hi @antinescience, I ended implementing this in a slightly different manner. However, and rightly so, I've added you to the [honour roll](https://github.com/CollaboratingPlatypus/PetaPoco/blob/development/honourRoll.md) for this contribution.

PR #174, opened 2013-12-10
Added support for DBNull in ExecuteScalar

Chri-s:
If SqlCommand.ExecuteScalar() returns DBNull.Value and the return type T of ExecuteScalar<T> is nullable, an InvalidCastException is thrown.
This patch handles DBNull.Value the same way as null.

This happens for example if ExecuteScalar<int?>("SELECT MAX(id) FROM empty_table") is queried on an empty table.


pleb:
Thanks for contributing! Keep up the good work

PR #193, opened 2015-01-15
Fix pagination

diegodfsd:
Fix count when use distinct statement with multiple columns and change sqlserverce ConnectionString to use DataDirectory.


pleb:

PR #249, opened 2016-02-03
Additional query support and a null reference guard

wis3guy:
Two minor changes:

1. added support for queries with custom sql, starting with the `WITH` or `SET` keyword;
2. added a null reference guard to the propertyinfo in the AddParam method;

### DIFF
--- a/PetaPoco/Database.cs
+++ b/PetaPoco/Database.cs
@@ -430,7 +430,7 @@ namespace PetaPoco
             {
                 p.Value = DBNull.Value;
 
-                if (pi.PropertyType.Name == "Byte[]")
+                if ((pi != null) && (pi.PropertyType.Name == "Byte[]"))
                 {
                     p.DbType = DbType.Binary;
                 }
--- a/PetaPoco/Utilities/AutoSelectHelper.cs
+++ b/PetaPoco/Utilities/AutoSelectHelper.cs
@@ -12,7 +12,7 @@ namespace PetaPoco.Internal
 {
     internal static class AutoSelectHelper
     {
-        private static Regex rxSelect = new Regex(@"\A\s*(SELECT|EXECUTE|CALL)\s",
+        private static Regex rxSelect = new Regex(@"\A\s*(SELECT|EXECUTE|CALL|WITH|SET)\s",
             RegexOptions.Compiled | RegexOptions.Singleline | RegexOptions.IgnoreCase | RegexOptions.Multiline);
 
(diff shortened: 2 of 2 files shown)


pleb:
Thanks @wis3guy for the contribute. 

Do you know which DBs have support for WITH and SET? I've never used them and I'd like to add a few integration tests.

Also, with that null check, do you know under what condition pi could be null (I'm thinking dynamic)? Again, I'd like to add an integration test.

wis3guy:
`WITH`
I know for sure that SQL Server supports it, that is what i need it for. See: https://msdn.microsoft.com/en-us/library/ms175972.aspx I also found this (http://www.brighthub.com/internet/web-development/articles/91893.aspx) *old* blog post, which states that Oracle 9i was the first to support it.

`SET`
Is used in T-SQL for variable assignment. See: https://msdn.microsoft.com/en-us/library/ms189484.aspx Afaik it is not used in the SQL language in that same context. There it is only used as a part of an `UPDATE` clause.

We ran into a problem where we would pass a single DTO object into the `Query<T>` method as the `args`. All properties were inspected and were translated into proper including the ones with `null` values. The `PropertyInfo` object `pi` however was also `null` for the properties with a `null` value.  If need be, i can probably provide a failing test somewhere beginning next week.

pleb:
`WITH`. Yep got. I'll add a test for SQL Server. Support will fall through to any other supported server.

`SET`. Of course. Face palm. 

Null check. I think I know what you mean. I'll see if I can get a test failing, if not I'll maybe call on that unit test.

pleb:
#250

pleb:
#251

pleb:
Hey @wis3guy. I've created two branches for each issues #250 and #251. In these branches I've added integration tests to confirm the bugs you reported. This way we can A) confirm the fix and B) confirm that they can't be introduced again.

Branches: [development_issue250](https://github.com/CollaboratingPlatypus/PetaPoco/tree/development_issue250) and [development_issue251](https://github.com/CollaboratingPlatypus/PetaPoco/tree/development_issue251)

For #250, can you please also include `DECLARE`, so `|WITH|SET|DECLARE`?

Lastly, sorry to be a pain, can you please issue two pull requests to the branches that I created.

wis3guy:
@pleb ok, so you want me to replay the changes on the new branches?

pleb:
@wis3guy yes please, if it's not too much trouble. I could of course add the code, but then there would be no record of your contribute ;)

wis3guy:
@pleb will do

PR #272, opened 2016-03-22
Using a clone to prevent an error if the param is a SqlParameter, 

dcrafti:
...rather than a value type.

The error is "The SqlParameter is already contained by another SqlParameterCollection"

### DIFF
--- a/PetaPoco/Utilities/ParametersHelper.cs
+++ b/PetaPoco/Utilities/ParametersHelper.cs
@@ -68,7 +68,12 @@ namespace PetaPoco.Internal
                 }
                 else
                 {
-                    args_dest.Add(arg_val);
+                    var argValCopy = arg_val;
+                    if (argValCopy is ICloneable)
+                    {   // Using a clone prevents the error "The SqlParameter is already contained by another SqlParameterCollection" if the parameter is a SqlParameter, rather than a value type.
+                        argValCopy = ((ICloneable)argValCopy).Clone();
+                    }
+                    args_dest.Add(argValCopy);
                     return "@" + (args_dest.Count - 1).ToString();
                 }
             }


pleb:
Hi @dcrafti, thanks for the contribute.

You're missing unit/integration tests to confirm the problem and the fix.

Also, if I understand this correctly, you're reusing SqlParameters in the multiple calls to PetaPoco? If so would the clone call be better placed in your code?

dcrafti:
The issue that the code fixes is that if there is sql passed into PetaPoco
like:
"select a from b where x != @0 and x like '@0%'" (for example), then
PetaPoco, under the hood, converts that to "select a from b where x != @0
and x like '@1%'" and it fills in @1 by duplicating @0.
That's usually fine, but it interacts poorly when using SqlParameters (and
possibly other things).
Here's why: A SqlParameter (or other IDbParameter) is sometimes needed when
doing a query by an indexable varchar. It's needed in that situation
because just passing through a string as a parameter to PetaPoco will have
the string treated as an nvarchar, which will cause the index to not be
used.
The expansion, by default, results in @0 and @1 being references to the
same parameter. Adding the exact same parameter to the sql twice will fail
because the parameter has already been used.
As such, cloning the parameter, when it is cloneable, can be done to ensure
that it is always a different parameter being used.

If you want to replicate the broken behaviour, try running something like
this:
_db.Fetch<IdDTO>("select 1 from [table] where @0 = 'test' and @0 != 'test'
", new SqlParameter { DbType = DbType.AnsiString, Value = "test" })
It doesn't matter about what's in the SQL. It only matters that one
IDbParameter is used twice.

pleb:
No worries. When I get time I'll investigate and write a few integration tests the prove the bug

kurda-kurda:
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using System.Data;
using System.Data.SqlClient;
using System.Windows.Forms;

namespace superkk.Data
{
    class ClsAccess
    {
        SqlConnection con;

        public ClsAccess()
        {
            con = new SqlConnection(@"Server = .\SQLEXPRESS; database = SuperHH; Integrated Security = true");
        }
        public void Open()
        {
            if (con.State != ConnectionState.Open)
            {
                con.Open();
                MessageBox.Show("پەیوندیەکە سەرکەتوبوو ", "Hamza IT", MessageBoxButtons.OK, MessageBoxIcon.Information);

            }

        }
        public void Close()
        {
            if (con.State == ConnectionState.Closed)
            {
                con.Close();
                MessageBox.Show("بەداخەوەهیچ پەیوەندیەک نی یە ", "Hamza IT", MessageBoxButtons.OK, MessageBoxIcon.Exclamation);
            }

        }
        public void ExecuteCommand(string Stored_Procedure, SqlParameter[] param)
        {
            try
            {
                SqlCommand cmd = new SqlCommand();
                cmd.CommandType = CommandType.StoredProcedure;
                cmd.CommandText = Stored_Procedure;
                cmd.Connection = con;
                if (param != null)
                {
                    cmd.Parameters.AddRange(param);
                }
                cmd.ExecuteNonQuery();

            }
            catch
            {
                MessageBox.Show(" پەیوەندیەکە هەڵەیە < ", "Hamza IT", MessageBoxButtons.OK, MessageBoxIcon.Error);
                return;
            }
        }
        public DataTable SelectData(string Stored_Procedure, SqlParameter[] param)
        {

            SqlCommand cmd = new SqlCommand();
            cmd.CommandType = CommandType.StoredProcedure;
            cmd.CommandText = Stored_Procedure;
            cmd.Connection =con;

            if (param != null)
            {
                for (int i = 0; i < param.Length; i++)
                {
                  cmd.Parameters.Add(param[1]);
                }

            }
            SqlDataAdapter da = new SqlDataAdapter(cmd);
            DataTable dt = new DataTable();
            da.Fill(dt);
            return dt;

            }
      

            }

        }
    
        
    

      hello there  , 
I have issue with this parameter 
that prevent me to run my csharp c#
that say sqlparameter is already contained by another sqlparameterCollection 
please help me for this

asherber:
@kurda-kurda I'm going to delete your comment, because it's not related to PetaPoco. If you have a question about general `SqlParameter` use, I suggest that you ask it on [StackOverflow](https://stackoverflow.com). I also suggest that you simplify your example and make sure it has all the relevant code. You've provided an entire class, which is probably not necessary, but you don't show how you're calling the class or where the exception is thrown.

Issue #39, opened 2011-06-01
Feature suggestion: multiple result set support


litera:
> *I'm sorry if I write feature suggestion as an issue but I don't know of a better place to post a suggestion.*

It would be nice for PetaPoco to support multi result-set queries (at least in terms of stored procedures). This is the best way to avoid the **N+1** problem. Related properties (as One- or Multi- relations) could as well be automatically resolved either by using some sort of *relator* lambda expressions or POCO attributes. The latter being favoured so we don't have to repeat our code every time we'd like to auto-relate entities.

pleb:
Closing as @LankyPaddy has added this feature

PR #274, opened 2016-03-23
Multi result-set support

LankyPaddy:
Support for Sql query results with multiple results sets.

Simple/limited integration tests provided


LankyPaddy:
Addresses logged issue #39

PetaPoco.Tests.Integration/Models/OrderLine.cs
```diff
@@ -25,7 +25,10 @@ public class OrderLine
         public decimal SellPrice { get; set; }
 
         [ResultColumn]
-        public decimal Total { get; set; }
```
pleb: @LankyPaddy I need to confirm this isn't being used by other tests. Looks good so far

PR #278, opened 2016-03-29
add Set clause support in Sql class

u0hz:
add `Set` clause support in Sql class in `Sql` class that is almost the same like how to use `Where`

### DIFF
--- a/PetaPoco.Tests.Unit/Core/SqlTests.cs
+++ b/PetaPoco.Tests.Unit/Core/SqlTests.cs
@@ -191,6 +191,44 @@ namespace PetaPoco.Tests.Unit.Core
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


pleb:
@u0hz thanks for the contribute.

pleb:
@u0hz The only concern here is that people may use this and not parameters. Does this open PetaPoco for easier SQL injection attacks? I know with the use of strings, it's already possible, but newcomers may fail to use parameters and may default to this. Thoughts?

u0hz:
Yes, it does have the problem.
Although a `Set(string field,  object value)` function can avoid the problem above, but it's not so in common use. 
And on the other hand, `Where` may have the same problem. So I think the docs should make that point clear, and encourage people to use parameters instead of simply join strings.

pleb:
Good points ;)

PR #280, opened 2016-04-11
fixed type conversion for non-int backed enums, and better nullable handling

AtsushiSuzuki:
fix for https://github.com/CollaboratingPlatypus/PetaPoco/issues/197
and for https://github.com/CollaboratingPlatypus/PetaPoco/issues/153

Issue #197, opened 2015-05-07
enum Byte exception


fastlink30:
i have problem with enumeration, if i try to get data from class like this:

....
        public DateTime? ALLADATA { get; set; } // datetime (0)
        public string CODIPIVA { get; set; } // nvarchar (11)
        public enuAlbergoXme HOTELXME { get; set; } // tinyint (0)
    }

and enuAlbergoXme is like this

```
public enum enuAlbergoXme : int
{
    NESSUNO = 0,
    A,
    B,
    C,
    D,
    E
}
```

is all ok, but if i declare enum as byte, i can't get data

Eccezione first-chance di tipo 'System.InvalidCastException' in Peta.dll
System.InvalidCastException: Cast specificato non valido.
   in petapoco_factory_0(IDataReader )
   in PetaPoco.Database.<Query>d__2`1.MoveNext() in c:\GIT\XME_C\Peta\Models\PetaPoco.cs:riga 884
SELECT \* FROM xm_rappo WHERE k_unique = @0
     -> @0 [Int32] = "442662"

is bug?

Ste1io:
Duplicate of #281

Issue #153, opened 2013-02-18
Allow Nullable types for SQL Server


kpconnell:
I understand a custom mapper could be used but out of the box support for nullable CLR types, e.g. Double? Int? etc would be great.  They currently cause a casting exception.

herebebeasties:
I'm behind a firewall that won't let me submit a proper pull request, but you need to change it like so:

    - converter = delegate(object src) { return Convert.ChangeType(src, dstType, null); };
    + var underlyingType = Nullable.GetUnderlyingType(dstType) ?? dstType;
    + converter = src => Convert.ChangeType(src, underlyingType, null);

This is currently failing for things like coercing a `float` column into a `double?` data type.

pleb:
Closing as there's integration tests for nullable types now, and I don't think this is an issue anymore. Please reopen if you believe I'm wrong.

PR #281, opened 2016-04-12
re-issued PR for #280

AtsushiSuzuki:
re-issued PR for #280 
- fix type conversion for non-int backed enums - #197 
- fix type conversion for nullable types - #153


AtsushiSuzuki:
thanks.

PR #283, opened 2016-04-29
Produce POCOs in alphabetical order for diffing generated classes.

iadaz:
This is a change I've found very useful for my own source code control, when using PetaPoco with a very large database that comes from a third party supplier. I found that when the supplier changes their database schema, I couldn't easily see what they had changed because PetaPoco was generating the POCOs in a different order each time. Thus diffing old and new files wasn't usually helpful.

I've amended the TT core include to set the output order of database schema items so that generated classes come out in alphabetical order each time the classes are regenerated. This change applies to MS-SQL only, I haven't got access to the other platforms to make the same change for those.

I guess this might be a contentious change as it is likely to change the output order of the generated POCOs for those users who haven't already got this problem, for instance those with small databases. As such I won't be offended if the PR is rejected!


knunery:
This looks like a good change to me.

Issue #285, opened 2016-05-03
No support for converting Guid's to Strings on Query or Fetch


rufio620:
It appears that PetaPoco does not support converting Sql Server uniqueidentifer columns to strings when doing a fetch. The library throws _Object must implement IConvertible_. While I realize Guid does not implement IConvertable, this is a pretty common scenario and is supported by many other ORMs. 

Looking at the code in PocoData.cs on line 362 it appears a check for the opposite (string => guid) is beingdone. It seems like the there should also be another if clause to support this scenario where the database column is a unique identifier (Guid) and the destination is a string. This will make deaiing with SQL Server uniqueidentifer columns much more manageable.

PR #286, opened 2016-05-03
Fix for guid to string conversion issue; #285

groogiam:
Proposed fix for issue I submitted earlier today [https://github.com/CollaboratingPlatypus/PetaPoco/issues/285](url)

This is pretty straight forward just adds explicit conversion from Guid to string. Guid does not implement IConvertible so the else clause currently results in an exception. This is mostly applicable to Sql Server where uniqueidentifiers are mapped to guids by SqlClient but are often much easier to treat as strings.


pleb:
Hi @rufio620 thanks for the contribute. There needs to be an integration test before this can be merged. I don't mind doing it, but it'll have wait till I get time. Else you can update the PR with a test

PR #288, opened 2016-05-16
Fixing byte based enums

mjohn:
If your enum is based on byte type, peta poco fails on casting. JCKodel's solution works well, 
https://github.com/CollaboratingPlatypus/PetaPoco/issues/126


pleb:
Hi @mjohn thanks for the contribute. There needs to be an integration test before this can be merged. I don't mind doing it, but it'll have wait till I get time. Else you can update the PR with a test

PR #295, opened 2016-07-06
Pynej patch 1

pynej:
This is a simple addition of a Exclude by Prefix variable that can be used to supress a block of tables from being generated.  

It may work better to add this as a Regex instead but the above was all my project neeed.
Eg: 
``ExcludeTablesRegex = "^(cms|CMS|umbraco).*$";``
Or maybe even a ovverride to included only matched tables instead of defaulting to all.
``IncludeTablesRegex = "myPrefix.*";``

### DIFF
--- a/PetaPoco/T4 Templates/Database.tt
+++ b/PetaPoco/T4 Templates/Database.tt
@@ -10,6 +10,7 @@
 	ClassPrefix = "";
 	ClassSuffix = "";
 	TrackModifiedColumns = false;
+	ExcludePrefix = new string[] {}; // Exclude tables by prefix.
 
     // Read schema
 	var tables = LoadTables();
--- a/PetaPoco/T4 Templates/PetaPoco.Core.ttinclude
+++ b/PetaPoco/T4 Templates/PetaPoco.Core.ttinclude
@@ -90,6 +90,7 @@ bool GenerateOperations = false;
 bool GenerateCommon = true;
 bool GeneratePocos = true;
 bool TrackModifiedColumns = false;
+string[] ExcludePrefix = new string[] {};
 
 
 public class Table
@@ -418,6 +419,10 @@ Tables LoadTables()
 					result.RemoveAt(i);
 					continue;
 				}
+				if(StartsWithAny(result[i].ClassName, ExcludePrefix)) {
+					result.RemoveAt(i);
+					continue;
+				}
 			}
 
 			conn.Close();
@@ -453,6 +458,14 @@ Tables LoadTables()
 	}
 
         
+}
+
+bool StartsWithAny(string s, IEnumerable<string> items)
+{
+    if (s == null)
+        return false;
+
+    return items.Any(i => s.StartsWith(i));
 }
 
 abstract class SchemaReader


pleb:
Thanks for the PR! Looks good to me

well0549:
I really don't know what you did but after this change code generation really fails.
When i click on "Run custom tool" to regenerate the templates it immediately crashes.

I do not use the ExcludePrefix

Is there anything I can do to prevent this crash ?

Just to make clear what i did....

I am using 5.1.156-beta in another project. And that works.
Now I created a new solution with two projects..
One project with the latest Nuget Package and one with 5.1.156-beta.

Then i generated the code for the beta, that worked.
Then i Copied the Database.TT and the app.config from the working version to the new package. Just to make sure there were no types.

Then when i hit generate, i get the following in my Database.cs
ErrorGeneratingOutput

That's it....

If i copy the PetaPoco.Generator.ttinclude from the beta to the latest version same result

But if I copy the PetaPoco.Core.ttinclude i CAN generate....

So i did a compare between the TT files, and all that was changed is the changes you made.....

pynej:
Not really sure why it wouldn't work in your case.  The Error List in VS should show more details down lower. (Ie not looking for errors on the Databse.cs file, but rather on the .ttinclude files.

pynej:
Not really sure why it wouldn't work in your case. The Error List in VS should show more details down lower. (Ie not looking for errors on the Databse.cs file, but rather on the .ttinclude files.

pynej:
Not really sure why it wouldn't work in your case. The Error List in VS should show more details down lower. (Ie not looking for errors on the Databse.cs file, but rather on the .ttinclude files.

well0549:
I get the following errors

Error	6	Compiling transformation: Type expected	c:\Projects\ClassLibrary2\ClassLibrary1\Models\Generated\PetaPoco.Core.ttinclude	1	12	ClassLibrary1
Error	5	Compiling transformation: Method must have a return type	c:\Projects\ClassLibrary2\ClassLibrary1\Models\Generated\PetaPoco.Core.ttinclude	1	6	ClassLibrary1
Error	4	Compiling transformation: Invalid token 'this' in class, struct, or interface member declaration	c:\Projects\ClassLibrary2\ClassLibrary1\Models\Generated\PetaPoco.Core.ttinclude	1	1	ClassLibrary1
Error	1	A namespace cannot directly contain members such as fields or methods	C:\Projects\ClassLibrary2\ClassLibrary1\Models\Generated\Database.cs	1	1	ClassLibrary1

well0549:
Ha i figured it out.......

there is one cr/lf too much in PetaPoco.Core.ttinclude.
It is at the end.

If i remove that extra cr/lf it works.
(The extra was probably added by some editor)

pleb:
@well0549 @pynej any chance of a PR to fix this?

well0549:
Well,

I really don't know how github works and what procedures i have to follow.
+ I really don't have the time to fix it.

(Leaving for holliday tonite :-) )

pynej:
Lol. I made a pr to fix the other one. I'll post another for this.

pynej:
New PR #301

well0549:
Thanks

Op 14 jul. 2016 7:53 p.m. schreef "Jeremy Pyne" <notifications@github.com>:

PR #298, opened 2016-07-11
Pynej patch 2

pynej:
Another patch, this adds a ExplicitColumns option defaulting to on to maintain the current functionality.  If toggled off then the ExplicitColumns attribute is removed and Column attributes only added if the property name and column name don't match.


pleb:
@pynej changes look. Thanks for the contribute. I'm unable to merge these changes as github is saying there's a conflict. Can you merge with the development branch and update the PR thanks?

pleb:
Again, awesome work @pynej

PR #304, opened 2016-07-19
Update PetaPoco.Core.ttinclude to support Guid auto-generation in SQL

Kenstea:
Fixed auto-increment for Guid key when its default value is set in SQL server as newsequentialid() or newid().

Here is the case:
Insert a record into a table with primary key is Guid and its default value is newsquentialid() or newid().

Fix for the issue #305


pleb:
@Kenstea looks good. Thanks for the contribute!

Issue #319, opened 2016-08-03
PostGreSqlSchemaReader does not map uuid to System.Guid


jokokko:
PostGreSqlSchemaReader when switching on column's udt_name maps uuid to string (the default case). It should be mapped to System.Guid (https://www.postgresql.org/docs/9.0/static/datatype-uuid.html).

PR #320, opened 2016-08-03
Map uuid udt_name to System.Guid in PostGreSqlSchemaReader #319

jokokko:


pleb:
@jokokko thanks for the contribute ;)

PR #321, opened 2016-08-03
Exclude virtual columns from Oracle

nillis:


pleb:
@nillis thanks for the contribute ;)

PR #359, opened 2017-01-04
Add Async support

md-frank:
Add Async support

### DIFF
--- a/PetaPoco/DatabaseExtensions.cs
+++ b/PetaPoco/DatabaseExtensions.cs
@@ -13,22 +13,22 @@ namespace PetaPoco
 
         public static Task<int> ExecuteAsync(this IExecute execute, string sql, params object[] args)
         {
-            return Task.Run(() => execute.Execute(sql, args));
+            return Task.Factory.StartNew(() => execute.Execute(sql, args));
         }
 
         public static Task<int> ExecuteAsync(this IExecute execute, Sql sql)
         {
-            return Task.Run(() => execute.Execute(sql));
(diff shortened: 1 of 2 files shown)


pleb:
Hi @md-frank, 

Thanks for the PR. However, this isn't how async should be implemented. It should flow through the db provider. 

Thoughts?

md-frank:
For example IExecute.Execute, This is a synchronous method.
The same, Asynchronous version of Execute should be under the IExecute

md-frank:
Hi pleb,
Excuse me for any mistakes since my not-so-good English.

For example IExecute.Execute, This is a synchronous method.
The same, Asynchronous version of Execute should be under the IExecute

pflannery:
Even if `IExecute.ExecuteAsync` is using the asynchronous Task<T> object the synchronous will still problem occur in the database IO calls.

What's really needed for true asynchronous state is that all of the database providers offer their asynchronous IO methods.

I.E. [DbCommand.ExecuteScalarAsync](https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbcommand.executescalarasync?view=netframework-4.7#System_Data_Common_DbCommand_ExecuteScalarAsync) should be used instead of `DbCommand.ExecuteScalar` when a user wants an asynchronous workflow.

pleb:
@md-frank, thanks for the contribution. 

I'm going to close this because I've added full async support to PP. What I mean by this, is that the async I've added flows all the way to the DB Provider, and hopefully, forms a non-blocking routine, but that is in the hand of the actual provider. 

Thanks. If you need async support, it's currently in the beta build

PR #374, opened 2017-03-13
added some needed {} in if blocks

travislaborde:
https://trello.com/c/bt7lgXOG/4-if-without-is-a-danger-to-future-development

this is my first open source contribution so I wanted to find something very safe to start with :)

### DIFF
--- a/PetaPoco/Database.cs
+++ b/PetaPoco/Database.cs
@@ -49,7 +49,9 @@ namespace PetaPoco
         public Database()
         {
             if (ConfigurationManager.ConnectionStrings.Count == 0)
+            {
                 throw new InvalidOperationException("One or more connection strings must be registered to use the no paramater constructor");
+            }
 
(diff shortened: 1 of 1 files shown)


pleb:
@travislaborde congrats on your first contribute. However, adjusting the code style probably isn't a great place to start. For instance, I love single line IF statements. Now not everybody does, and I accept that, but I'm not going to merge this. Sorry. A better place to start would be to read the code to get an understanding of how it works, then review the list of issue people have reported, and see if you can't fix something.

travislaborde:
Wade, thanks for the advice.  I'll bow out of course, but I'd just like to
say that I don't think this is a style issue at all.  It is a danger to
future development.

Here is a link describing why:

https://trello.com/c/bt7lgXOG/4-if-without-is-a-danger-to-future-development

Thanks again for taking the time to respond.  I really appreciate it.  I
love PetaPoco!  Best of luck going forward,
Travis

PR #386, opened 2017-04-28
Situation for OracleConnection when using the Oracle.DataAccess.dll

CV80:
I working with PetaPoco in my project, this project using Oracle.DataAccess.dll for connection to Oracle DB, I fixed for the error under this situation.

### DIFF
--- a/PetaPoco/Core/DatabaseProvider.cs
+++ b/PetaPoco/Core/DatabaseProvider.cs
@@ -83,7 +83,7 @@ namespace PetaPoco.Core
         public virtual object MapParameterValue(object value)
         {
             if (value is bool)
-                return ((bool) value) ? 1 : 0;
+                return ((bool)value) ? 1 : 0;
 
             return value;
         }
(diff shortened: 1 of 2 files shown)


pleb:
I see why you did it that way. However, we has Oracle code leaking into the base class.

What about

        /// <summary>
        ///     Returns the .net standard conforming DbProviderFactory.
        /// </summary>
        /// <param name="assemblyQualifiedNames">The assembly qualified names of the provider factory.</param>
        /// <returns>The db provider factory.</returns>
        /// <exception cref="ArgumentException">Thrown when <paramref name="assemblyQualifiedName" /> does not match a type.</exception>
        protected DbProviderFactory GetFactory(params string[] assemblyQualifiedNames)
        {
            Type ft = null;

            foreach(var aqn in assemblyQualifiedNames)
            {
                ft = Type.GetType(aqn);

                if (ft != null)
                    break;
            }

            if (ft == null)
                throw new ArgumentException("Could not load the " + GetType().Name + " DbProviderFactory.");

            return (DbProviderFactory) ft.GetField("Instance").GetValue(null);
        }

and

        public override DbProviderFactory GetFactory()
        {
            return GetFactory("Oracle.ManagedDataAccess.Client.OracleClientFactory, Oracle.ManagedDataAccess", "Oracle.DataAccess.Client.OracleClientFactory, Oracle.DataAccess");
        }

Thoughts?

CV80:
Yes, it is better move child provider code to their child class.

PR #397, opened 2017-05-22
Fixed a bug that causes the column of the additional [ResultColumn] f…

TiomTang:
Fixed a bug that causes the column of the additional [ResultColumn] feature not to be retrieved in the query when the "select * from tableName" is omitted. The method involved: Query < T >,


TiomTang:
The values of the columns are not available

pleb:
Hi @TiomTang. Thanks for the contribute. 

A few things.

A) You need to PR to the development branch.

B) I have a feeling this will break something. Did you run the integration tests?

C) Can you add an integration test that covers this situation?

Thanks. Wade

asherber:
This behavior was by design, but #474 added the possibility of including result columns in auto select queries.

PR #399, opened 2017-05-23
The Columns of the [ResultColumn] caused by "pd. Columns" cannot be c…

TiomTang:
The Columns of the [ResultColumn] caused by "pd. Columns" cannot be correctly valued.sorry,I don't understand English very well, maybe the meaning is not clear enough.


asherber:
This behavior was by design, but #474 added the possibility of including result columns in an autoselect query.

PR #403, opened 2017-06-29
Create TeradataProvider.cs

nathanbaker:
Adding support for Teradata

### DIFF
--- a/PetaPoco/Providers/TeradataProvider.cs
+++ b/PetaPoco/Providers/TeradataProvider.cs
@@ -1,0 +1,53 @@
+using System;
+using System.Data;
+using System.Data.Common;
+using PetaPoco.Core;
+using PetaPoco.Internal;
+using PetaPoco.Utilities;
+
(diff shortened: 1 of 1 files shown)


pleb:
Thanks for the PR. Is there a docker image for this DB? The reason I ask, is because I'd like to figure out how easy/hard it would be to add integration tests

nathanbaker:
That's a good idea, it seems this page has docker-images for Teradata

https://github.com/Teradata/docker-images

I can have a look into it and get back to you.

pleb:
Sorry it's been so long. I've looked and they still don't have docker images. All I could find was a VMware image on the TeraData website. The TeraData website is a nightmare to navigate. Then I tried to find the documentation to read for TeraData database itself. To cut a long story short, I've given up. I'm going to leave this PR here encase somebody comes along and can make use of it. I really only want to support DBs I can run via docker so that I can run integration tests against them

PR #408, opened 2017-07-21
Support for value converter attribute

### LINKED ISSUE #401: Add support for column value converters
Add attribute for custom column value converter. This is useful for storing .Net types to database, such as  `Image` as VARBINARY, `Color` as INT32. For example:

`[ImageValueConverter] public Image Icon { get; set; }`
`[ColorValueConverter] public Color ForeColor { get; set; }`

This could be done by simple modification of `ConventionMapper` default constructor and 
abstract `ValueConverterAttribute`.

See [this gist](https://gist.github.com/tkouba/1b7ff5a06c353b0bfdff3a17aee291a9) for code example and usage.

tkouba:
This closes #401 and add simple support for value converter defined by attribute. Real value converters is not included, so integration test was not added.


pleb:
Thanks for the PR @tkouba 

My only concern is that this will only work with the convention mapping method. Not a big issue, however, I'd like a few days to see if we can't implement in the old mapping methods. I'll keep the #401 open until I can get some time to look at it. 

The pain points for introducing into the old mapping method will be modifying the IL generated code. The Umbraco devs have told me that the IL gen code can be removed without much of an impact on pref. Maybe it's time. 

I'll merge the PR which will put it dev and the pre-release nuget channel.

PR #432, opened 2018-01-08
Enable .NETStandard 2.0 (.NETCoreApp 2.0 and .NETFramework 4.0-4.7.1)

soeleman:
This pull consist new solution (PetaPoco.2017) so people who still using VS2015 still able continue as usual.
VS2017.3 needed because support of .NETCoreApp 2.0 dan .NETStandard 2.0, specially for multi-targeting.

The .NETStandard 2.0, still using original source code of PetaPoco. We only make 'link as' and make specific target with Define (e.g. for .NETFramework as general we using NETFULL). The changes is minimal only on DataBase.cs and additional define on level project.  

Tested on Test.Unit and Test.Integration both work as expected. On some scenario have some error because difference drives.


pleb:
@soeleman v6 supports .net stanard. thanks for the help 😄

PR #453, opened 2018-05-01
Some change for Access mdb

vcatrambone:
I wrote 
1.   class **OleDbSchemaReader : SchemaReader** in T4 template PetaPoco.Core.ttinclude to read the table information for Access DB ( views are skipped)

2) on **PetaPoco/Database.cs** a particular case per DateTime, because Access DB not accept millisecond in his field

3) **PetaPoco/Core/PocoData.cs** **PetaPoco/Core/MultiPocoFactory.cs**
particular case for Access DB, in case of table join that have a fields with same name Access add name of table before the field name

### DIFF
--- a/PetaPoco/Core/MultiPocoFactory.cs
+++ b/PetaPoco/Core/MultiPocoFactory.cs
@@ -91,7 +91,12 @@ namespace PetaPoco.Internal
             {
                 // Split if field name has already been used, or if the field doesn't exist in current poco but does in the next
                 var fieldName = r.GetName(pos);
-                if (usedColumns.ContainsKey(fieldName) || (!pdThis.Columns.ContainsKey(fieldName) && pdNext.Columns.ContainsKey(fieldName)))
+                var pdThisWithTableName = pdThis.Columns.FirstOrDefault(x => pdThis.TableInfo.TableName + "." + x.Key == fieldName);
+                var pdNextWithTableName = pdNext.Columns.FirstOrDefault(x => pdNext.TableInfo.TableName + "." + x.Key == fieldName);
+
+                if (usedColumns.ContainsKey(fieldName) || (!pdThis.Columns.ContainsKey(fieldName) && pdNext.Columns.ContainsKey(fieldName))
+                    || (pdThisWithTableName.Equals(new KeyValuePair<string, PocoColumn>()) && !pdNextWithTableName.Equals(new KeyValuePair<string, PocoColumn>()))
+                 )
                 {
                     return pdThis.GetFactory(sql, connectionString, firstColumn, pos - firstColumn, r, defaultMapper);
                 }
--- a/PetaPoco/Core/PocoData.cs
+++ b/PetaPoco/Core/PocoData.cs
@@ -227,7 +227,17 @@ namespace PetaPoco.Core
                         // Get the PocoColumn for this db column, ignore if not known
                         PocoColumn pc;
                         if (!Columns.TryGetValue(reader.GetName(i), out pc))
-                            continue;
+                        {
+                            var k = Columns.FirstOrDefault(x => TableInfo.TableName + "." + x.Key == reader.GetName(i));
+                            if (!k.Equals(new KeyValuePair<string, PocoColumn>()))
+                            {
+                                pc = k.Value;
+                            }
+                            else
+                            {
+                                continue;
+                            }
+                        }
 
                         // Get the source type for this column
                         var srcType = reader.GetFieldType(i);
--- a/PetaPoco/Database.cs
+++ b/PetaPoco/Database.cs
@@ -488,6 +488,27 @@ namespace PetaPoco
                     p.GetType().GetProperty("UdtTypeName").SetValue(p, "geometry", null); //geography is the equivalent SQL Server Type
                     p.Value = value;
                 }
+                else if (value.GetType().Name.Contains("DateTime") && Provider.GetType() == Type.GetType("PetaPoco.MsAccessDbDatabaseProvider"))
+                {
+                    var copy = new DateTime(
+                          ((DateTime)value).Year
+                        , ((DateTime)value).Month
+                        , ((DateTime)value).Day
+                        , ((DateTime)value).Hour
+                        , ((DateTime)value).Minute
+                        , ((DateTime)value).Second
+                        , 0 //Millisencond 
+                            //0= Works
+                            //1= Exception in insert and update
+                        , ((DateTime)value).Kind
+                        );
+                    //var k = new DateTime(1000, 1, 1, 1, 1, 1, DateTimeKind.Local);
+                    //var copy = (DateTime)value;
+
+                    //value = (copy).AddMilliseconds(-copy.Millisecond);
+
+                    p.Value = copy;
+                }
                 else
                 {
                     p.Value = value;
--- a/PetaPoco/T4 Templates/PetaPoco.Core.ttinclude
+++ b/PetaPoco/T4 Templates/PetaPoco.Core.ttinclude
@@ -454,6 +454,11 @@ Tables LoadTables()
 				// Oracle
 				reader=new OracleSchemaReader();
 			}
+			else if(_factory.GetType().Name == "OleDbFactory")
+			{
+			  reader =new OleDbSchemaReader ();
+			} 
+
 			else
 			{
 				// Assume SQL Server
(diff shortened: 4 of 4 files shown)


PetaPoco/Core/MultiPocoFactory.cs
```diff
@@ -90,8 +90,13 @@ private static Delegate FindSplitPoint(Type typeThis, Type typeNext, string conn
             for (; pos < r.FieldCount; pos++)
             {
                 // Split if field name has already been used, or if the field doesn't exist in current poco but does in the next
-                var fieldName = r.GetName(pos);
-                if (usedColumns.ContainsKey(fieldName) || (!pdThis.Columns.ContainsKey(fieldName) && pdNext.Columns.ContainsKey(fieldName)))
+                var fieldName = r.GetName(pos);
+                var pdThisWithTableName = pdThis.Columns.FirstOrDefault(x => pdThis.TableInfo.TableName + "." + x.Key == fieldName);
```
pleb: Can you explain what this is for? It's been a while since I've looked at MultiPocoFactory

PetaPoco/Core/MultiPocoFactory.cs
```diff
@@ -90,8 +90,13 @@ private static Delegate FindSplitPoint(Type typeThis, Type typeNext, string conn
...
             {
                 // Split if field name has already been used, or if the field doesn't exist in current poco but does in the next
-                var fieldName = r.GetName(pos);
-                if (usedColumns.ContainsKey(fieldName) || (!pdThis.Columns.ContainsKey(fieldName) && pdNext.Columns.ContainsKey(fieldName)))
+                var fieldName = r.GetName(pos);
+                var pdThisWithTableName = pdThis.Columns.FirstOrDefault(x => pdThis.TableInfo.TableName + "." + x.Key == fieldName);
+                var pdNextWithTableName = pdNext.Columns.FirstOrDefault(x => pdNext.TableInfo.TableName + "." + x.Key == fieldName);
+
+                if (usedColumns.ContainsKey(fieldName) || (!pdThis.Columns.ContainsKey(fieldName) && pdNext.Columns.ContainsKey(fieldName))
+                    || (pdThisWithTableName.Equals(new KeyValuePair<string, PocoColumn>()) && !pdNextWithTableName.Equals(new KeyValuePair<string, PocoColumn>()))
```
pleb: This `pdThisWithTableName.Equals(new KeyValuePair<string, PocoColumn>()` seems a little off.  I'm wondering if you simply need `pdThisWithTableName != null`

PetaPoco/Database.cs
```diff
@@ -488,6 +488,27 @@ private void AddParam(IDbCommand cmd, object value, PropertyInfo pi)
...
+                        , ((DateTime)value).Day
+                        , ((DateTime)value).Hour
+                        , ((DateTime)value).Minute
+                        , ((DateTime)value).Second
+                        , 0 //Millisencond 
+                            //0= Works
+                            //1= Exception in insert and update
+                        , ((DateTime)value).Kind
+                        );
+                    //var k = new DateTime(1000, 1, 1, 1, 1, 1, DateTimeKind.Local);
```
pleb: Can you please remove the commented out code

PetaPoco/Core/MultiPocoFactory.cs
pleb: Actually, sorry, I think I see what you're doing. Maybe this should be `pdThisWithTableName == typeof( KeyValuePair<string, PocoColumn>)`

PetaPoco/Core/PocoData.cs
```diff
@@ -225,9 +225,19 @@ public Delegate GetFactory(string sql, string connectionString, int firstColumn,
                     for (int i = firstColumn; i < firstColumn + countColumns; i++)
                     {
                         // Get the PocoColumn for this db column, ignore if not known
-                        PocoColumn pc;
-                        if (!Columns.TryGetValue(reader.GetName(i), out pc))
-                            continue;
+                        PocoColumn pc;
+                        if (!Columns.TryGetValue(reader.GetName(i), out pc))
+                        {
+                            var k = Columns.FirstOrDefault(x => TableInfo.TableName + "." + x.Key == reader.GetName(i));
```
pleb: A dictionary TryGetValue approaches an O(1) operation. If the column doesn't exist, we might be adding significant overhead here, as this would be O(N). Obviously, you added it for a reason. Can I ask why you're checking for `KeyValuePair<string, PocoColumn>` in here and the multi poco? I can't see you adding this, so I'm thinking maybe it was missing?

PetaPoco/Core/MultiPocoFactory.cs
vcatrambone: **pdThisWithTableName != null**
it make build error: Operator '!=' cannot be applied to operands of type 'KeyValuePair<string, PocoColumn>' and '<'null'>'	
 var pdThisWithTableName = pdThis.Columns.FirstOr**Default**(x => pdThis.TableInfo.TableName + "." + x.Key == fieldName);
**Default** is: "new KeyValuePair<string, PocoColumn>()" not 'null'

(pd**This**WithTableName.Equals(new KeyValuePair<string, PocoColumn>()) 
&& ***!*** pd**Next**WithTableName.Equals(new KeyValuePair<string, PocoColumn>()))

I think that i wrote is correct, but i check again..

PetaPoco/Core/PocoData.cs
vcatrambone: AccessDB  in case of joins between tables can return this format
TableName1.exampleName, TableName2.exampleName
and  for example: 
reader.GetName(i) =="TableName1.exampleName" or 
reader.GetName(i) =="TableName2.exampleName"

PetaPoco/Core/MultiPocoFactory.cs
asherber: `KeyValuePair` is a value type and is not nullable, but I think what you want is `default(KeyValuePair<string, PocoColumn>)` rather than `new`.

vcatrambone: I think that "default(KeyValuePair<string, PocoColumn>)" is elegant but make the same result of 
new KeyValuePair<string, PocoColumn>() :)

PetaPoco/Core/PocoData.cs
pleb: Ah, I see

PetaPoco/Core/MultiPocoFactory.cs
pleb: Ah, I see.

So I guess it comes down to these two

    pdThisWithTableName.GetType() == typeof(KeyValuePair<string, PocoColumn>)

or

    pdThisWithTableName == default(KeyValuePair<string, PocoColumn>)
    pdThisWithTableName == new KeyValuePair<string, PocoColumn>()

I think we should choose the one which has better performance.

*Thinking out loud*

Can either be optimised by (reduce allocations & etc)

    var t = typeof(KeyValuePair<string, PocoColumn>);
    pdThisWithTableName.GetType() == t

    var i = default(KeyValuePair<string, PocoColumn>);
    var i = new KeyValuePair<string, PocoColumn>();
    pdThisWithTableName == new KeyValuePair<string, PocoColumn>()

Maybe even

    private static KeyValuePair<string, PocoColumn> kvpType = typeof(KeyValuePair<string,PocoColumn>);
    private static kvpInstance = default(KeyValuePair<string, PocoColumn>);

vcatrambone: Pleb, i'm not the best developer but:
1. pdThisWithTableName always is  KeyValuePair<string, PocoColumn>
 ```
pdThisWithTableName.GetType() == typeof(KeyValuePair<string, PocoColumn>)
```

2. It is impossibile a build error
```
pdThisWithTableName == new KeyValuePair<string, PocoColumn>() 
```
3.
```
private static KeyValuePair<string, PocoColumn> kvpType = typeof(KeyValuePair<string,PocoColumn>);
````
it is not consistent...
 
I think  that best performance can be  
```
pdThisWithTableName.Key==null
```
in 
```
if (usedColumns.ContainsKey(fieldName) || (!pdThis.Columns.ContainsKey(fieldName) && pdNext.Columns.ContainsKey(fieldName))
                    || ( /*CHANGE with pdThisWithTableName.Key==null*/pdThisWithTableName.Equals(new KeyValuePair<string, PocoColumn>()) && !pdNextWithTableName.Equals(new KeyValuePair<string, PocoColumn>()))
```
I hardly believe that much can still be done..

PR #457, opened 2018-05-22
add query method

maybgit:
add query method  return datatable

### DIFF
--- a/PetaPoco/Database.cs
+++ b/PetaPoco/Database.cs
@@ -1030,7 +1030,7 @@ namespace PetaPoco
             OpenSharedConnection();
             try
             {
-                DataTable table = new DataTable();
+                var table = new DataTable();
                 using (var cmd = CreateCommand(_sharedConnection, sql, args))
                 {
                     try
--- a/PetaPoco/Database.cs
+++ b/PetaPoco/Database.cs
@@ -1033,18 +1033,15 @@ namespace PetaPoco
                 var table = new DataTable();
                 using (var cmd = CreateCommand(_sharedConnection, sql, args))
                 {
-                    try
-                    {
-                        table.Load(cmd.ExecuteReader());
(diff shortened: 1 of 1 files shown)


asherber:
I don't think this is a good idea. The whole point of an ORM is to return strongly typed objects. If you want a DataTable, there's no reason to use PetaPoco.

pleb:
Yeah I can’t see why you would need petapoco to use a data table. @maybgit is there a reason why you need this over the normal method of creating a dt in .net?

maybgit:
@pleb @asherber Thank you for your reply
I would like to say my thoughts, or do you have any good suggestions for my need to use DataTable?

1.DataTable can be used for SqlBulkCopy.WriteToServer, such as synchronization from MySql to SqlServer, or between different databases.
2. some API only need to simply get DataTable serialization return, do not need update or insert or other operations, do not need too much Poco Entity, do not think that repeated conversion is a good way
3. I think ORM+DataAccess=PetaPoco, for example, want to get only count (*), you can use Single<T> or SingleOrDefault<T>, you need to get Pocl Entity for OO development, you can use Query<T>, you need DataTable, and you can use Query.
4. I hope PetaPoco will become more flexible and suitable for different development needs.

pleb:
I feel this is probably better as an extension for your own project

    public static class MyPetaPocoExtensions
    {
        public static DataTable Query(this Database pp, string sql, params object[] args)
        {
            pp.OpenSharedConnection();
            try
            {
                var table = new DataTable();
                using (var cmd = pp.CreateCommand(pp.Connection, sql, args))
                    table.Load(cmd.ExecuteReader());
                return table;
            }
            finally
                pp.CloseSharedConnection();
        }
    }

I believe that most users wouldn't need this functionality

maybgit:
ok thanks. @pleb

PR #459, opened 2018-06-01
Convert to multitargeted .NET Standard/Framework project

asherber:
I'd love to get this library .NET Standard compatible. Maybe we can use this PR to work on it.

 It looks like the only needed code change is `GetFactory()` in the MSAccess provider. It also means upgrading to VS 2017. I'm happy to help with multi-targeting the unit tests, but I'm not sure I have the right setup for the integration tests.

### DIFF
--- a/PetaPoco.sln
+++ b/PetaPoco.sln
@@ -1,13 +1,8 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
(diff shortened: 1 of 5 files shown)


pleb:
I've added a branch **netstandard**. @asherber can you target that branch with this PR?

Let's track this work in a ticket. https://github.com/CollaboratingPlatypus/PetaPoco/issues/460

asherber:
New PR is #462

PR #463, opened 2018-06-01
Remove System.Configuration

asherber:
Following the example of NPoco, I removed (for .NET Standard) Database constructors that assumed app.config/web.config and made appropriate changes to the fluent constructor.


asherber:
I also made the unit tests project multi-targeted. Tests are passing for both Framework and Standard.

Issue #421, opened 2017-10-20
Feature request - IgnoreOnInsert and IgnoreOnUpdate properties


Andrei-:
I often have the scenario where I have a table column like CreatedDate or UniqueGuid column which should be handled by the database only. So basically I never want to touch those columns when inserting or updating pocos, but I want the values when I query.
I couldn't solve this nicely with ResultColumn or Ignore attributes as they will not load the column. If there is an elegant solution and I missed the obvious, I would love to hear it..

What I ended up doing was having two classes for each table. For example on entities that need a CreatedDate, I have a class without that property for inserting and updating and a second class with that property for querying. The second class is just inheriting the first and adds the CreatedDate property.
Any chance of having these attributes added into Petapoco?

asherber:
The docs are a little misleading wrt `ResultColumn`. They say:

> ResultColumn is an attribute which can decorate a Poco property as a result only column. A result only column is a column that is only populated in queries and is not used for updates or inserts operations.

The important thing that's missing here is that a `ResultColumn` is **not included in auto-generated SQL**. So if you do `var records = db.Fetch<MyClass>("")`, result columns will not be populated. But if you do `var records = db.Fetch<MyClass>("select * from myclass")`, result columns will be populated. 

NPoco adds a third attribute, `ComputedColumn`, which is kind of nice. Like `ResultColumn`, a computed column won't be included in an insert or update, but unlike `ResultColumn` it will be included in an auto-generated SELECT statement.

Andrei-:
Coming from Petapoco, I didn't notice the new ComputedColumn feature. When researching npoco, I read that it adds a bunch of extra features, but could never find a detailed list of exactly what new features npoco adds over Petapoco. 
ComputedColumn seems to be the answer to my problem, so thanks a lot, I really appreciate you took the time to answer here.

iadaz:
Could I request that at least the wiki is updated with Andrei's comment "ResultColumn is not included in auto-generated SQL". This tripped me up.

I worked around this by extending the ResultColumn attribute with an IncludeInAutoSelect flag as suggested here: https://stackoverflow.com/a/9939919/1499294

Ultimately this ends up modifying the QueryColumns collection in PocoData():

` QueryColumns = (from c in Columns where !c.Value.ResultColumn || c.Value.IncludeInAutoSelect select c.Key).ToArray();`

pleb:
[...]

PR #474, opened 2018-09-22
Include result column in auto select

asherber:
I decided to implement this with an enum instead of a bool param in the constructor -- `ResultColumn(false)` just looked too much like it was saying that the field is _not_ a result column.

In `ColumnInfo` and `PocoColumn`, I called the field `AutoSelectedResultColumn`, because keeping the name as `IncludeInAutoSelect` made its purpose sound too broad.


asherber:
Addresses #421

asherber:
@pleb Do you know what this AppVeyor error is?

pleb:
Away for a few hours. Will look when I get home

Issue #400, opened 2017-05-27
Query error in Database.Single with EnableAutoSelect = False


MohanVijayakumar:
It Seems when call the DB.Single<T>(primarykey)  with  EnableAutoSelect = false Petapoco only Generates Where Part.

pleb:
@MohanVijayakumar what do you think the behaviour should be under this situation? Throw?

MohanVijayakumar:
It would be great if it generates the full query
The same issue found in the petapoco fork of Npoco also
[https://github.com/schotime/NPoco/issues/183](url)

PR #475, opened 2018-09-26
Generate a SELECT for Single<T> even if EnableAutoSelect is false

asherber:
Addresses #400 

This could use a couple of integration tests.


asherber:
AppVeyor again

Issue #479, opened 2018-10-22
Parameterless overloads for Fetch and Query


asherber:
I've always felt that `Fetch()` and `Query()` should have parameterless overloads, equivalent to `Fetch("")`. I'm happy to prepare a PR if @pleb thinks this is appropriate.

pleb:
Sounds good, as it doesn't break existing APIs and only adds to API surface. PR away ;)

PR #480, opened 2018-10-29
Add overloads without sql param

asherber:
Give Fetch(), Query(), Page(), and SkipTake() overloads that omit the
string sql parameter, so that PetaPoco will do SELECT *

It didn't seem to make sense to add overloads for the multi-poco versions, so I just did the base versions.


asherber:
Closes #479

PR #481, opened 2018-10-30
Teach IsNew about decimal primary keys

### LINKED ISSUE #466: IsNew method doesn't consider Decimal field.
When Save is called on POCO, it calls IsNew method to check if the PrimaryKey field contains default value of the datatype. IsNew method though considers every other possible datatypes for the column, but has no reference to DECIMAL datatype. If the PrimaryKey field in SQL Server is of type Numeric, it causes PetaPOCO to not properly identify if the record is new or not & thus calls UPDATE instead of INSERT.

asherber:
Closes #466 

I'm not sure if this is desirable, because I'm not sure what kind of table has a decimal primary key. But it doesn't seem to do any harm.


asherber:
Whoops, branched from the wrong place. I'll try again.

PR #482, opened 2018-10-30
Teach IsNew() about decimal primary key

### LINKED ISSUE #466: IsNew method doesn't consider Decimal field.

asherber:
Not sure if this desirable, since I don't know what kind of table would have a decimal PK, but I don't think it would do any harm.

Closes #466


pleb:
The AppVeyor build fails are killing me. Can't wait for github actions ™️, so we can remove it.

asherber:
Is there something I can do with my pull requests to make things build properly? What's the issue?

pleb:
I don't think so. There's some weird going on with AppVeyor and Powershell. I would investigate further, but it seems intermittent and I want to replace it with something else. GH actions look promising. I've signed up for the beta. Hopefully, that accept my application 😄

PR #487, opened 2018-11-06
Stored procs

### LINKED ISSUE #486: Support for stored procs
I am working on adding first-class support for stored procs. The methods I am currently adding are: 

```csharp
IEnumerable<T> QueryProc<T>(string storedProcedureName, params object[] args);

List<T> FetchProc<T>(string storedProcedureName, params object[] args);

T ExecuteScalarProc<T>(string storedProcedureName, params object[] args);

int ExecuteNonQueryProc(string storedProcedureName, params object[] args);
```

Because I am implementing these just like existing methods (working with an `IDbCommand` and adding named parameters), the `args` must be either POCOs (each readable property will be turned into a named parameter) or pre-constructed `IDbDataParameter`s. Passing any value types or strings will throw an exception. 

I'm very interested in any feedback on the design and usability as I work on this.

asherber:
Closes #486 

I'm not 100% sold on the names of these methods, so please let me know if you have other ideas.

One oddity in the Firebird ADO.NET provider is that while it appears to support named parameters for stored procs, it actually just sends them in order, without regard for the name. So anyone using stored procs with PetaPoco will need to be careful to pass things in order and not just assume that the names will sort everything out. See https://www.tabsoverspaces.com/227021-firebird-net-provider-and-calling-stored-procedures-with-parameters

I also had to make a change to the way the Sql Server integration tests run. You had a single script which ran under the `mssql_builder` connection, so that the db could be created if it didn't exist. The problem I ran into was that there was no way, using this method, to create the test stored procs in the PetaPoco db. (Couldn't have a `USE` statement, because `CREATE PROCEDURE` needs to be the first thing in a batch -- and in any case, the script was getting split at each GO.) So now the db creation is handled in code (`MssqlDBTestProvider.cs`), and then the whole script is run under the `mssql` connection, which has PetaPoco as the initial catalog.

Anyway, please let me know if you have any suggestions regarding design or implementation.


pleb:
Re `Firebird ADO.NET`, yeah weird. There's not much we can do about it though. Maybe we need a doco page which lists provider quirks.

Issue #488, opened 2018-11-20
Ad Hoc querying with table valued parameters?


strangepholker:
We've got a need to perform queries that include long lists of ids...potentially thousands of ids.  Our initial implementation was a simple where IN (@0) implementation and the list of ids was passed in as an argument.  

e.g.
.Where($"Id IN (@0)", itemIdBatchParameter)

However, this approach has a couple of problems.  First, with a sufficiently large list of items it's possible to exceed SQL server's 2100 maximum parameter count.  We've dealt with that by creating batches of ids and making multiple queries and then concatenating the resulting collections of POCOs.  However, we also suspect this use of the IN with a highly variable number of parameters is causing problems with the SQL Server query optimizer as a separate query plan is generated for each permutation in the parameter count.  

Ideally, I'd like to join to a table valued parameter.  The raw SQL for doing this is provided below. 

```
-- Declare User-Defined Table type.
-- This would be defined outside of the scope of a given query
Create Type [medical].GuidList as Table
(
	Id uniqueidentifier primary key
)

-- User Defined Table type is populated with 'id' values from a previous query
Declare @VersionIds medical.GuidList
Insert into @VersionIds values ('FE3D8792-4D86-4AB6-9023-0ED321C79B24')
Insert into @VersionIds values ('9DAC2440-3ACB-4FCA-9AC4-6AFF73ABDF9F')
Insert into @VersionIds values ('BB2E7C00-7BA6-487E-8F02-C0A20C082851')

-- Filtering is performed by joining to the table valued parameter
-- rather than using WHERE DocumentVersionId IN ('FE3D8792-4D86-4AB6-9023-0ED321C79B24', '9DAC2440-3ACB-4FCA-9AC4-6AFF73ABDF9F', 'BB2E7C00-7BA6-487E-8F02-C0A20C082851', ...)
select * from medical.[DocumentSetEntry] setEntry
inner join @VersionIds versions
ON setEntry.[DocumentVersionId] = versions.Id

```

I'm able to declare the SqlParameter in c# like this:

```
protected SqlParameter CreateIdListParameter(IEnumerable<Guid> ids, string parameterName)
{
	const string columnName = "Id";
	var table = new DataTable();
	table.Columns.Add(columnName, typeof(Guid));

	foreach (var guid in ids)
	{
		var row = table.NewRow();
		row[columnName] = guid;
		table.Rows.Add(row);
	}

[...]

PR #490, opened 2018-11-21
Allow for an IDbDataParameter to be added directly to Sql

asherber:
I'm not sure how you feel about this. It helps with the issue in #488, but it may be unnecessary or may lead to confusion or misuse. I don't feel strongly about arguing for its inclusion.

### DIFF
--- a/PetaPoco.Tests.Unit/Core/SqlTests.cs
+++ b/PetaPoco.Tests.Unit/Core/SqlTests.cs
@@ -5,6 +5,7 @@
 // <date>2018/07/02</date>
 
 using System;
+using System.Data.SqlClient;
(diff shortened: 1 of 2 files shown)


pleb:
Yeah. Hmm. I can see people being confused about what it's used for.

pleb:
Actually, it should be ok and I see it has a purpose.

asherber:
Sorry, this isn't working correctly -- I didn't follow it all the way through. The param is getting added to the Sql, but it's not making it through `CreateCommand()` and getting added to the `IDbCommand`. I'll work on fixing this and get you an updated PR.

asherber:
@pleb This needs further discussion. Should we do that here, or open a new issue?

PR #497, opened 2018-12-13
Add generic db constructor, and add optional mapper to all

asherber:
When I've been writing little tests for things in LinqPad, I've been using the `Database` constructor that takes a connection string and a provider name, and I've been a little frustrated that I can't always remember the exact name/spelling of a provider. I know that I could use `nameof(SqlServerDatabaseProvider)` to get help from Intellisense, but I thought it would be simpler and more idiomatic if I could just do `new Database<SqlServerDatabaseProvider>(connString)`.

While I was in there, I made sure that all `Database` constructors have the optional `IMapper` parameter, because I couldn't see why some had it and some didn't.

### DIFF
--- a/PetaPoco.Tests.Unit/DatabaseTests.cs
+++ b/PetaPoco.Tests.Unit/DatabaseTests.cs
@@ -11,7 +11,9 @@ using System.Data.SqlClient;
 using System.Dynamic;
 using Moq;
 using PetaPoco.Core;
+using PetaPoco.Providers;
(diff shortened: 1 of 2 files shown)


iadaz:
Sorry accidentally clicked on the request review link. 

I had this very same issue yesterday so I think this is a good idea.

asherber:
@pleb I'd appreciate it if you could look at this and #499 when you have a chance. These are things I have a need for, and I'd rather use an official build than a private one.

Issue #485, opened 2018-11-01
Version 6.1 - Add Async support


pleb:
The recent work by @asherber has motivated me to devote a little more time to PetaPoco. Thanks, @asherber for chipping in a doing such a great job!

Over the next month or so I'm going to tackle Async support. I'm assuming, like .net standard, this is a feature that people would "just expect" to be there when they open the box. 

I plan to follow the MS guidance -> [Task-based Asynchronous Pattern (TAP)](https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap).

At this stage, I'll be applying the [configure await false pattern](https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f) for library code. 

Any feedback is welcome.

tomekjaworski:
Great news! 😃

asherber:
See #141; maybe work with @tmenier or build on what he's done with [AsyncPoco](https://github.com/tmenier/AsyncPoco)?

samchenws:
Great news! And thanks for your awesome contribution 😃

tmenier:
You're bumping into all the pain points I had one upon a time. :)

asherber:
1. AsyncPoco and NPoco have both implemented `QueryAsync()` -- is there something that can be gleaned from their approaches?

2. This feels funny to me -- it doesn't seem like the way that sync/async code is usually implemented. I haven't looked at it in detail, but I would vote for keeping the implementations separate. Yes, there would be some duplicated code, but I don't think there would be much; seems like you'd mostly need two copies of `ExecuteReader()` and its brethren.

pleb:
Re point 1. There are ways around the issue. However, it will cause a differences between the async verse sync API. The workarounds end up being not as clean. For example, with AsyncPoco, the workaround is a callback the caller passes in. I haven't looked at NPoco, but I assume it'll be the same. I'd rather wait for c# 8 before async is added for this particular API. 

Re point 2. The issue is that async and sync code paths are not compatible. In c# the compiler writes out a statemachine for any await/task usage. So the choice is either duplicate code or favour async. I'll keep an eye out, but I haven't seen any nice solutions to date for this problem.

asherber:
[...]

PR #507, opened 2019-02-24
Add async support

pleb:


pleb:
Tag #485

asherber:
Wondering if a pattern like this might be useful to reduce code duplication: https://gist.github.com/asherber/8c5eaab9975e393de6ed752a0e30a1cb

See the Flag Argument Hack at https://msdn.microsoft.com/en-us/magazine/mt238404.aspx

pleb:
I had some similar, but .net 4 doesn’t support the async method modifier. So the use of #if becomes an ugly mess

asherber:
Interesting, I wasn't getting a compiler error on that, but when I made a new test app in .NET 4 just to reproduce some of this, it wouldn't run.

Take a look at the gist now.  Not the prettiest, but I suspect it's better than duplicating all that code.

pleb:
At this stage, it's likely that only the connection and query method will be duplicated. We can revisit the connection one later on if we need to. The query will likely always be duplicated, as I could not find a way of making yield work with async. That said, when we can roll c# 8 we can revisit it again, but it probably going to be something that can't be changed due to the language restrictions we have now and that fact it would be a breaking change later on. I looked at how dapper got around it, and well, they just disable the streaming that yield gives and return a pre-populated list, which is effectively the PP fetch method.

PR #519, opened 2019-04-19
Some cleanup on XML docs

asherber:


pleb:
Great work. I think most of those typos are mine. Thanks, for fixing them. 💪

asherber:
Ah, no worries. These things sneak in over time, and sometimes you just have to sit down and fix them.

PR #530, opened 2019-05-11
Refactor to get rid of duplicate TableInfo/ColumnInfo code

asherber:
@pleb Interested in your thoughts on this. I noticed that while `StandardMapper` calls methods in `ColumnInfo` and `TableInfo` to get information, `ConventionMapper` duplicates that code and then adds a little extra processing (like inflection). I thought it would be good to refactor so that `ConventionMapper` can call the same base methods in `ColumnInfo` and `TableInfo` and then do what it needs to.

(One practical implication of the duplication is that when I added `IncludeInAutoSelect` to `ResultColumnAttribute`, I only handled it it `ConventionMapper`, because I didn't see the other methods. Now it's handled in `ColumnInfo` and picked up by both `ConventionMapper` and `StandardMapper`.)

The refactored methods are not as clean as I would like them – they've got a couple of `out` params so that `ConventionMapper` can avoid a second round of reflection. On the whole, that's probably an okay tradeoff.

### DIFF
--- a/PetaPoco/Core/ColumnInfo.cs
+++ b/PetaPoco/Core/ColumnInfo.cs
@@ -1,4 +1,6 @@
-﻿using System.Reflection;
+﻿using System;
+using System.Linq;
+using System.Reflection;
 
 namespace PetaPoco
 {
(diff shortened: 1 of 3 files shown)


pleb:
I'll have a better look tonight. But, from your description and logical conclusion, I'd say you've likely picked the best route. There are a few parts which could do with a big refactor tidy, namely the mapping (this) and ILGen mapper generator.

pleb:
Sorry, I keep getting help up at night.

I've had a read through, and everything seems to line up.

Issue #538, opened 2019-05-24
IsNew always return false when PrimaryKey is typeof string and value is ""


ArgoZhang:
hi, I found that IsNew function always return false when the poco's PrimaryKey is System.String type and value is "".
in line 2532 will return. so line 2535 will be never work.
```
if (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>) || !type.IsValueType)
    return pk == null;

if (type == typeof(string))
    return string.IsNullOrEmpty((string) pk);
```
it will work fine when the poco PrimaryId value is null.
is it by design? or a bug?

ArgoZhang:
how about this? any question?

ArgoZhang:
still busy? really hope resolve IsNew and Mapper.Register issues. So many string type Primary key in my project. i do not want to create a new class inherit Database and override IsNew function. any question please let me known. thanks

pleb:
@ArgoZhang , this is now in dev. Use the latest beta nuget package to try it out

ArgoZhang:
@pleb I've tested it. It's working very well.

PR #542, opened 2019-05-31
refactor(IsNew): check string type PrimaryKey first #538

ArgoZhang:
#Issue
link https://github.com/CollaboratingPlatypus/PetaPoco/issues/538


pleb:
Hey @ArgoZhang why was this causing an issue? I don't see how this change affects anything.

ArgoZhang:
> Hey @ArgoZhang why was this causing an issue? I don't see how this change affects anything.

```
[TableName("Groups")]
public class Group
{
    public string Id { get; set; }
    public string GroupName { get; set; }

    public void Save(Group poco)
    {
        var db = DbManager.Create();
        db.Save(new Group() { Id = "", GroupName = "UnitTest" });
    }
}
```
because Id value is **String.Empty** not **null**, so IsNew method not work correctly.

pleb:
@ArgoZhang could you please fix the line endings?

ArgoZhang:
@pleb done! resolved the conflicts.
thank you for pick up this PR.
😃

PR #552, opened 2019-09-26
Update should return 0 when columns is empty

### LINKED ISSUE #551: Save method crashes
if you call this method for not modified item or not new item, it throws an exception. Since  there is no modified columns, there is no column name in SET part of the produced SQL statement.  I think if an entity is not new or no modification done on it, Save method should just do nothing? Right? Or am I mistaken?

asherber:
Fixes #551 

@pleb I can also backport the changes to v5 if you want.


pleb:
I wouldn't worry about backporting ;)

asherber:
Should I note in the original issue that the fix won't be backported, and the OP can upgrade to v6 and still use his generated files?

pleb:
@asherber I've replied to OP

Issue #562, opened 2019-11-24
byte[], Binary, PreExecute


asherber:
I'm working with an oddball database engine for which I wrote my own custom `DatabaseProvider`. This database has a `Blob` field type, which I can read just fine using a `byte[]` property. But when I try to write to a table with one of those fields, I get `ArgumentException: Unable to infer parameter data type`. This has led me to a couple of questions.

1. **Should `byte[]` automatically map to `DbType.Binary`?** There is no code in `Database.SetParameterProperties()` to handle `byte[]` -- except when the parameter value is null and the property type is `byte[]`, in which case the `DbType` does get set to `Binary`. It seems like always treating `byte[]` as `Binary` would be a good thing. (NPoco does this, FWIW.)

1. **Should `DatabaseProvider.PreExecute()` get called in a different place?** I initially worked around my issue by using the `Database.CommandExecuting` event and setting the `DbType` to `Binary` for any `byte[]` properties. Then I thought that since `SetParameterProperties()` doesn't handle this globally, I could use `PreExecute()` in my `DatabaseProvider` to accomplish the same thing. The help text for `PreExecute()` says "Called immediately before a command is executed", but in fact it's called as part of `Database.CreateCommand()`. This means that for Insert and Update operations, `PreExecute()` sees the `DbCommand` object before any SQL or parameters have been added, which isn't very useful. What about calling `_provider.PreExecute()` as part of `Database.DoPreExecute()`, which does get called as part of every db call?

I'm happy to work on PRs for these.

pleb:
Both sound good. Good for it ;)

PR #563, opened 2019-11-29
Force byte[] params to map to DbType.Binary

asherber:
Relates to #562 

This was trivial to implement, but I'm not sure if we should actually do it. After looking at some more DataProviders, I see that most of them handle this kind of translation from .NET type to DbType themselves. The particular db I'm working with for some reason doesn't explicitly handle `byte[]`, which is why I was having problems. But if I can count on `DatabaseProvider.PreExecute()` being called (see other PR), then I can handle this myself in my custom `DatabaseProvider`.

So I'm perfectly fine if you want to decline this PR.

### DIFF
--- a/PetaPoco.Tests.Unit/CreateCommandTests.cs
+++ b/PetaPoco.Tests.Unit/CreateCommandTests.cs
@@ -144,5 +144,17 @@ namespace PetaPoco.Tests.Unit
             Action act = () => _db.CreateCommand(_conn, CommandType.StoredProcedure, "procname", arg);
             act.ShouldThrow<ArgumentException>();
         }
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


pleb:
I'm happy to include it. I have a feeling this should be pretty safe and introduce no regression. Do you I've made the right assumption regarding this?

asherber:
I think it's probably safe, and most likely the right way to handle `byte[]` but I'm also hesitant to introduce a change that doesn't fix an obvious mistake or omission, and that nobody seems ever to have complained about before. I also wouldn't want to base a fix solely on a problem with a db that is not very widely used; I can now address the issue I was having by modifying my provider's `PreExecute()`, which I think is what that method is there for.

So on the whole, I don't see a compelling reason for this PR and would recommend declining it.

PR #580, opened 2020-05-15
KeyNotFoundException with concise message

nils-a:
catches KeyNotFoundException and throws InvalidOperationException containing the invalid value and the Enum-Name in which the value should have been mapped.

### DIFF
--- a/PetaPoco/Utilities/EnumMapper.cs
+++ b/PetaPoco/Utilities/EnumMapper.cs
@@ -27,10 +27,11 @@ namespace PetaPoco.Internal
             {
                 return map[value];
             }
-            catch (KeyNotFoundException)
+            catch (KeyNotFoundException inner)
             {
-                throw new InvalidOperationException(
-                    string.Format("Can not convert value `{0}` into {1}-Enum", value, enumType.Name));
+                throw new KeyNotFoundException(
+                    $"Requested value '{value}' was not found in enum {enumType.Name}.",
+                    inner);
             }
         }
     }
--- a/PetaPoco/Utilities/EnumMapper.cs
+++ b/PetaPoco/Utilities/EnumMapper.cs
@@ -23,7 +23,15 @@ namespace PetaPoco.Internal
                 return newmap;
             });
 
-            return map[value];
+            try
+            {
+                return map[value];
+            }
+            catch (KeyNotFoundException)
+            {
+                throw new InvalidOperationException(
+                    string.Format("Can not convert value `{0}` into {1}-Enum", value, enumType.Name));
+            }
         }
     }
 }
--- a/PetaPoco.Tests.Unit/Core/EnumMapperTest.cs
+++ b/PetaPoco.Tests.Unit/Core/EnumMapperTest.cs
@@ -1,0 +1,49 @@
+﻿using PetaPoco.Internal;
+using Shouldly;
+using System;
+using System.Threading.Tasks;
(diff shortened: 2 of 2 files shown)


PetaPoco/Utilities/EnumMapper.cs
```diff
@@ -23,7 +23,15 @@ public static object EnumFromString(Type enumType, string value)
...
 
-            return map[value];
+            try
+            {
+                return map[value];
+            }
+            catch (KeyNotFoundException)
+            {
+                throw new InvalidOperationException(
+                    string.Format("Can not convert value `{0}` into {1}-Enum", value, enumType.Name));
```
asherber, line 33: If I call `Enum.Parse()` with a bad value, I get an `ArgumentException` with message `Requested value 'foo' was not found.` I think the exception and message should be the same here -- or possibly append the type name as well.

```csharp
throw new ArgumentException($"Requested value '{value}' was not found in enum {enumType.Name}.");
```

(Note also that technically this is a breaking change. Maybe a compromise would be to keep the `KeyNotFoundException` and just change the message?)

nils-a: I like it. (I really didn't like the `InvalidOperationException` because that's not what it is.. I would have loved to have a `EitherYourDatabaseIsCorruptedOrYourPocosAreOutOfDateException` but one can't have it all, can one?)

I'll go with the `KeyNotFoundException` - that fits just as well and avoids the breaking change..

pleb:
Seems ok to me. @asherber ?

asherber:
I'm wondering a little why the code in `EnumMapper` is needed at all. Couldn't we just use `Enum.Parse(typeof(Foo), "MyValue", true)`?

Or to preserve the current exception type, as @nils-a did:

```csharp
static class EnumMapper
{
    public static object EnumFromString(Type enumType, string value)
    {
        try
        {
            return Enum.Parse(enumType, value, true);
        }
        catch (ArgumentException inner)
        {
            throw new KeyNotFoundException(
                    $"Requested value '{value}' was not found in enum {enumType.Name}.",
                    inner);
        }
    }
}
```

Does the cache code really improve performance? It looks like `Enum.Parse()` uses a cache internally as well.

nils-a:
@asherber Having done a little test, it seems the current implementation is actually a little slower than your proposed implementation.

I used [this code](https://gist.github.com/nils-a/5389d22810eb356f04adf2c276a62117#file-thebenchmark-cs) as a benchmark. Without errors (i.e. all strings are valid enum-names) the [outcome](https://gist.github.com/nils-a/5389d22810eb356f04adf2c276a62117#file-outcome-txt) looks like this:

```
|        Method |       Mean |    Error |   StdDev |
|-------------- |-----------:|---------:|---------:|
| UsingProposed |   826.2 ns | 26.99 ns | 79.15 ns |
| UsingExisting | 1,130.2 ns | 33.76 ns | 98.48 ns |
```

> Note: When adding a string that is not a valid name, the [outcome is different](https://gist.github.com/nils-a/5389d22810eb356f04adf2c276a62117#file-outcome-with-nonexistingname-txt).
> 
> Also note the test was hastily cobbled together and done on "my" machine, which does whatnot in the background, so it's probably not a "good" benchmark

This seems to point towards the proposed implementation: It's a bit faster and much more readable.

pleb:
Thanks for the PR 🤟

PR #612, opened 2021-07-06
iOS compatibility changes

6pac:
Add alternate non-IL Factory method to PocoData.cs. This is usable under iOS (under Xamarin), which blocks IL generation and some reflection operations.

These changes are a first iteration. They are tested on iOS, but use only the commonly used PetaPoco features.

Because the error is runtime and not compile time, we may wish to add a configuration option to allow use with either code path. This depends on if the non-iOS compatible path yields significant performance improvements (I would imagine this is why it was moved to IL).

Also, MultiPocoFactory.cs has a CreateMultiPocoFactory method (and perhaps GetAutoMapper) which would need to be converted for complete coverage. I've never used the MultiPoco feature before and thought I'd test the waters as to how essential that is before I leap into another half day of trying to back-convert IL.

### DIFF
--- a/PetaPoco/Core/PocoData.cs
+++ b/PetaPoco/Core/PocoData.cs
@@ -108,6 +108,86 @@ namespace PetaPoco.Core
             return tc >= TypeCode.SByte && tc <= TypeCode.UInt64;
         }
 
+        // convert a IDataReader record into a POCO
+        public T PocoFactory<T>(string sql, string connectionString, int firstColumn, int countColumns, IDataReader reader, IMapper defaultMapper) {
+            var mapper = Mappers.GetMapper(Type, defaultMapper);
+
+            T poco = Activator.CreateInstance<T>();
+            //T opoco= new T();
+
+            if (Type == typeof(object)) {
+                // Enumerate all fields generating a set assignment for the column
+                for (int i = firstColumn; i < firstColumn + countColumns; i++) {
+                    var srcType = reader.GetFieldType(i);
+                    var srcName = reader.GetName(i);
+                    var val = reader.GetValue(i);
+                    if (val == DBNull.Value) val = null;
+
+                    // Get the converter
+                    if (val != null) {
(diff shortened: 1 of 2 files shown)


6pac:
I note the check failure appears to have nothing to do with the code!

asherber:
I see that this commit has been reverted in the github-actions branch, but it's still available in development (and the beta package) and is broken.

Prior to this commit, [this line](https://github.com/CollaboratingPlatypus/PetaPoco/blob/1817f1f70bed769472ccd8a1c9bf25505513177a/PetaPoco/Database.cs#L1472) returned a (cached) function that could be called on an `IDataReader`. In this commit, the [same line](https://github.com/CollaboratingPlatypus/PetaPoco/blob/d1c2fed4888eeb22d32886fa348c8c9d1d94ae3b/PetaPoco/Database.cs#L1471) now _invokes_ the (non-cached) function on the `IDataReader`, which is not yet open. The result is an exception on every call to `Query()`. And even there were no exception, trying to cast the returned `T` to a `Func<IDataReader, T>` would probably yield null, which would throw an exception a few lines later when the function is used.

6pac:
Probably my bad, the code had changed quite a bit since my original fork, so I did my best at porting and hoped there was adequate unit test coverage to catch any problems. The testing is challenging to set up locally because of the various databases.

pleb:
Sorry, I've reverted this in the development branch too. The plan was to finish swapping over to GH Actions and pull the revert in as part of the merge.

@6pac I think this is close, but as @asherber said it needs a delegate wrapper.

Ideally, all the il generation would go in favour of delegates or expression trees. Though I have my doubts of il/express trees being that much faster than delegates. Having this part of the codebase readable might also encourage a bit of innovation, as I'm sure it scares off a fair few

6pac:
OK, as I've probably said, this is working in my version of PP which is 4.0.3. I don't have the time right now to update my codebase to the latest, and I probably won't have the time for a while to install, update and test the current version.  Will get to it when I can.

PR #621, opened 2021-08-22
Specify correct assembly name for SqlClientFactory

asherber:
This type lives in System.Data.SqlClient under .NET Core.  
Some systems seem to resolve it correctly with System.Data,  
but not all.


asherber:
Holding off on this for now. I thought it might have been a Windows thing, since my machine has both Framework and Core on it, but I'm able to run a console app in .NET Core on Linux and `System.Data` resolves correctly. I'd like to follow up with the Linqpad folks to see if there's a particular reason that it's not working the same there.

asherber:
This turned out to be an [issue with Linqpad](https://forum.linqpad.net/discussion/2502/system-data-sqlclient-types-not-found-in-lp6/p1?new=1) (more specifically, an issue with applications that dynamically load assemblies). I'm going to close this PR -- I think the approach is reasonable, but there's no reason to introduce it at this point.

PR #623, opened 2021-09-03
Allow dictionaries to be passed with parameters

asherber:
We can currently pass in POCOs that hold parameter values (for queries and stored procs) and PP will turn them into arrays of parameters. This PR lets us pass in Dictionaries as well.

```c#
db.Query(sql, new { Foo = 5, Bar = "apple" });

// same as 
var dict = new Dictionary<string, object>()
{
    ["Foo"] = 5,
    ["Bar"] = "apple"
}
db.Query(sql, dict);
```

The reason for this is that it lets you build up a collection of arbitrarily named parameters.


iadaz:
I thought this was useful enough to document. I've added the missing Wiki "Querying" page just by cobbling together examples from elsewhere, and added this at the bottom under "future dev". Hope that's OK.

Issue #188, opened 2014-11-06
Adding DateTime2 parameter 


maxoizs:
Adding full DateTime with milliseconds in petepoco `AddParam` will be treated as `DateTime` which will loss the milliseconds part, and if I amend the if else case with 

``` c#
else if (item.GetType().Name == "DateTime")
        {
          p.DbType = DbType.DateTime2;
          p.Value = item;
        }
```

will treat all `DateTime` parameters as `DateTime2` which ok in my case as I'm using `DateTime2` everywhere but it'll invalid the where clause for `DateTime` parameter if it's against `DateTime` only column. 

I wonder if it possible to determine the column type for the parameter ?

pleb:
Seems like something that PP should allow to be configured

0ndrej:
I just bumped into this problem now. Recommended type for MSSQL date is datetime2, while PetaPoco is mapping everything to datetime. If I create POCO class with .NET DateTime property, exception is thrown when trying to Insert, because of SQL datetime overflow even though it is defined as datetime2 in database.

tkouba:
One possible solution is adding function to interface IProvider `DbType MapParameterValue(object value, PropertyInfo pi);` and use it in Database.AddParam. What is your opinion?

maxoizs:
Sorry guys, It's been so long and I've actually forgot about it totally. therefore I'm sure I won't be of much help.

pleb:
No worries mate. I should have some free time starting mid-year. Tickets, list this one, are on my list ;)

roggmaeh:
I just have the same problem. Is a fix like AnsiString workaround (varchar/nvarchar) possible?

roggmaeh:
AnsiString and DateTime2 fix
https://github.com/roggmaeh/PetaPoco/commit/7ef1a8fb9409653e67a85f969f2b76b012d2b226

my table model
[Column("DATECHANGE", IsDateTime2 = true)]
public DateTime2 DATECHANGE {get;set;)

[Column("NAME", IsAnsi = true)]
public string NAME {get;set;}

asherber:
@roggmaeh Can you please open a pull request so we can review and comment?

asherber:
@roggmaeh Can you please open a pull request so we can review and comment?

asherber:
Please see my comments in the [PR](https://github.com/CollaboratingPlatypus/PetaPoco/pull/630#pullrequestreview-840961143). I'm not sure that this work is needed.

PR #630, opened 2021-12-27
ansi and datetime2 fix

roggmaeh:
the normal AnsiString fix is fine for Fetch<> ...
var test = db.FirstOrDefault<TABLE>("where NAME = @0", new AnsiString("YOURNAME"));

but for Update and Insert i cant use AnsiString 
test.SOMEDATETIME2COLUMN = DateTime.Now();
test.SOMESTRINGCOLUMN = "Hello";
test.Update() 

my new model looks like this
[Column("SOMEDATETIME2COLUMN", IsDateTime2 = true)] public DateTime? SOMEDATETIME2COLUMN { get; set;}
[Column("SOMESTRINGCOLUMN", IsAnsi = true)] public string SOMESTRINGCOLUMN { get; set;}

### DIFF
--- a/PetaPoco/Attributes/ColumnAttribute.cs
+++ b/PetaPoco/Attributes/ColumnAttribute.cs
@@ -38,11 +38,11 @@ namespace PetaPoco
         /// <summary>
         ///     True if AnsiString / VARCHAR column
         /// </summary>
-        public bool IsAnsi { get; set; }
+        public bool ForceToAnsiString { get; set; }
         /// <summary>
         ///     True if DateTime2 column
         /// </summary>
-        public bool IsDateTime2 { get; set; }
+        public bool ForceToDateTime2 { get; set; }
 
         /// <summary>
         ///     Constructs a new instance of the <seealso cref="ColumnAttribute" />.
--- a/PetaPoco/Attributes/ColumnAttribute.cs
+++ b/PetaPoco/Attributes/ColumnAttribute.cs
@@ -35,6 +35,14 @@ namespace PetaPoco
         ///     json)"
         /// </summary>
         public string UpdateTemplate { get; set; }
+        /// <summary>
+        ///     True if AnsiString / VARCHAR column
+        /// </summary>
+        public bool IsAnsi { get; set; }
+        /// <summary>
+        ///     True if DateTime2 column
+        /// </summary>
+        public bool IsDateTime2 { get; set; }
 
         /// <summary>
         ///     Constructs a new instance of the <seealso cref="ColumnAttribute" />.
--- a/PetaPoco/Core/DateTime2.cs
+++ b/PetaPoco/Core/DateTime2.cs
@@ -1,0 +1,22 @@
+﻿using System;
+
+namespace PetaPoco
+{
+    /// <summary>
+    ///     Wrap strings in an instance of this class to force use of DBType.DateTime2
+    /// </summary>
+    public class DateTime2
+    {
+        /// <summary>
+        ///     The DateTime value
+        /// </summary>
+        public DateTime Value { get; }
+
+        /// <summary>
+        ///     Constructs an DateTime2
+        /// </summary>
+        /// <param name="str">The C# string to be converted to ANSI before being passed to the DB</param>
+        public DateTime2(DateTime str)
+            => Value = str;
+    }
+}
--- a/PetaPoco/Core/DateTime2.cs
+++ b/PetaPoco/Core/DateTime2.cs
@@ -3,7 +3,7 @@
 namespace PetaPoco
 {
     /// <summary>
-    ///     Wrap strings in an instance of this class to force use of DBType.DateTime2
+    ///     Wrap DateTime in an instance of this class to force use of DBType.DateTime2
     /// </summary>
     public class DateTime2
     {
@@ -15,8 +15,10 @@ namespace PetaPoco
         /// <summary>
         ///     Constructs an DateTime2
         /// </summary>
-        /// <param name="str">The C# string to be converted to ANSI before being passed to the DB</param>
-        public DateTime2(DateTime str)
-            => Value = str;
+        /// <param name="dt">The C# DateTime to be converted to DateTime2 before being passed to the DB</param>
+        public DateTime2(DateTime dt)
+            => Value = dt;
+
+        public static explicit operator DateTime2(DateTime dt) => new DateTime2(dt);
     }
 }
--- a/PetaPoco/Core/DateTime2Extentions.cs
+++ b/PetaPoco/Core/DateTime2Extentions.cs
@@ -1,0 +1,14 @@
+﻿using System;
+
+namespace PetaPoco
+{
+    public static class DateTime2Extentions
+    {
+        /// <summary>
+        /// Converts an object to its <see cref="DateTime2"/> representation
+        /// </summary>
+        /// <param name="o">The object to be converted</param>
+        /// <returns></returns>
+        public static DateTime2 ToDateTime2(this object o) => new DateTime2((DateTime)o);
+    }
+}
--- a/PetaPoco/Database.cs
+++ b/PetaPoco/Database.cs
@@ -591,7 +591,7 @@ namespace PetaPoco
         /// <param name="cmd">A reference to the IDbCommand to which the parameter is to be added</param>
         /// <param name="value">The value to assign to the parameter</param>
         /// <param name="pi">Optional, a reference to the property info of the POCO property from which the value is coming.</param>
-        private void AddParam(IDbCommand cmd, object value, PropertyInfo pi)
+        private void AddParam(IDbCommand cmd, object value, PropertyInfo pi, PocoColumn pc)
         {
             // Convert value to from poco type to db type
             if (pi != null)
@@ -616,18 +616,14 @@ namespace PetaPoco
             {
                 var p = cmd.CreateParameter();
                 p.ParameterName = cmd.Parameters.Count.EnsureParamPrefix(_paramPrefix);
-                SetParameterProperties(p, value, pi);
+                SetParameterProperties(p, value, pi, pc);
 
                 cmd.Parameters.Add(p);
             }
         }
 
-        private void SetParameterProperties(IDbDataParameter p, object value, PropertyInfo pi)
+        private void SetParameterProperties(IDbDataParameter p, object value, PropertyInfo pi, PocoColumn pc)
         {
-            var columnAttribInfo = pi?.GetCustomAttributes(typeof(ColumnAttribute), true).FirstOrDefault() as ColumnAttribute;
-            var forceToAnsiString = columnAttribInfo?.ForceToAnsiString == true;
-            var forceToDateTime2 = columnAttribInfo?.ForceToDateTime2 == true;
-
             // Assign the parameter value
             if (value == null)
             {
@@ -643,12 +639,12 @@ namespace PetaPoco
 
                 var t = value.GetType();
 
-                if (t == typeof(string) && forceToAnsiString == true)
+                if (t == typeof(string) && pc?.ForceToAnsiString == true)
                 {
                     t = typeof(AnsiString);
                     value = value.ToAnsiString();
                 }
-                if (t == typeof(DateTime) && forceToDateTime2 == true)
+                if (t == typeof(DateTime) && pc?.ForceToDateTime2 == true)
                 {
                     t = typeof(DateTime2);
                     value = value.ToDateTime2();
@@ -753,7 +749,7 @@ namespace PetaPoco
             cmd.CommandText = sql;
 
             foreach (var item in args)
-                AddParam(cmd, item, null);
+                AddParam(cmd, item, null, null);
 
             return cmd;
         }
@@ -1899,7 +1895,7 @@ namespace PetaPoco
 
                 names.Add(_provider.EscapeSqlIdentifier(i.Key));
                 values.Add(string.Format(i.Value.InsertTemplate ?? "{0}{1}", _paramPrefix, index++));
-                AddParam(cmd, i.Value.GetValue(poco), i.Value.PropertyInfo);
+                AddParam(cmd, i.Value.GetValue(poco), i.Value.PropertyInfo, i.Value);
             }
 
             var outputClause = string.Empty;
@@ -2156,7 +2152,7 @@ namespace PetaPoco
                     sb.AppendFormat(i.Value.UpdateTemplate ?? "{0} = {1}{2}", _provider.EscapeSqlIdentifier(i.Key), _paramPrefix, index++);
 
                     // Store the parameter in the command
-                    AddParam(cmd, i.Value.GetValue(poco), i.Value.PropertyInfo);
+                    AddParam(cmd, i.Value.GetValue(poco), i.Value.PropertyInfo, i.Value);
                 }
             }
             else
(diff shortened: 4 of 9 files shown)


asherber:
I will review by tomorrow.

PetaPoco/Database.cs
```diff
@@ -673,6 +689,19 @@ private void SetParameterProperties(IDbDataParameter p, object value, PropertyIn
                     // Thanks @DataChomp for pointing out the SQL Server indexing performance hit of using wrong string type on varchar
                     p.DbType = DbType.AnsiString;
                 }
+                else if (t == typeof(DateTime2))
+                {
+                    var asValue = (value as DateTime2).Value;
+                    if (asValue == null)
```
asherber, line 695: This `if..then` block can be a single line:

```c#
p.Value = asValue ?? DBNull.Value;
```

PetaPoco/Database.cs
```diff
@@ -624,6 +624,10 @@ private void AddParam(IDbCommand cmd, object value, PropertyInfo pi)
 
         private void SetParameterProperties(IDbDataParameter p, object value, PropertyInfo pi)
         {
+            var columnAttribInfo = pi?.GetCustomAttributes(typeof(ColumnAttribute), true).FirstOrDefault() as ColumnAttribute;
```
asherber, line 627: This seems a little kludgy to me. My hunch is to say that what's really wanted is to get the Ansi and DT2 properties onto `PocoColumn`, and then change calls to `AddParam` and `SetParameterProperties` so that they pass in the `PocoColumn` instead of `PropertyInfo`. (There would still be access to the `PropertyInfo` via `PocoColumn.PropertyInfo`, but now we can also check Ansi and DT2 directly.) 

@pleb What do you think?

asherber (changes requested):
@roggmaeh I've added some comments for things that I think should be fixed or cleaned up. I also have some questions about the overall approach, but I defer to @pleb on this.

1. I'm not sure that there's an advantage to specifying `DbType.AnsiString` on inserts and updates. The issue really has to do with selects -- if the parameter is a different type from the index on the column, then the index can't be used. But I think on insert/update the type will just be coerced with no penalty.

2. I'm also not 100% sure that we need this for `DateTime2` either. The C# `DateTime` class only goes down to milliseconds, which means that any loss of precision will occur when _reading_ from a `DateTime2` field, not _writing_ to one. I don't know if there's the same issue with indexes that there is with strings on selects, but again I don't think we need to worry about writes.

If we do want to implement all or part of this, I also think that new functionality like this requires at least some unit tests, and probably some integration tests as well to prove things out.

asherber:
Closes #188

PetaPoco/Core/DateTime2.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace PetaPoco
+{
+    /// <summary>
+    ///     Wrap strings in an instance of this class to force use of DBType.DateTime2
```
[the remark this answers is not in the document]
roggmaeh: fixed in new commit

PetaPoco/Core/DateTime2.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <summary>
+        ///     The DateTime value
+        /// </summary>
+        public DateTime Value { get; }
+
+        /// <summary>
+        ///     Constructs an DateTime2
+        /// </summary>
+        /// <param name="str">The C# string to be converted to ANSI before being passed to the DB</param>
+        public DateTime2(DateTime str)
```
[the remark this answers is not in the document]
roggmaeh: fixed in new commit

PetaPoco/Core/DateTime2.cs
```diff
@@ -0,0 +1,22 @@
...
+            => Value = str;
+    }
```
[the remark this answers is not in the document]
roggmaeh: added in new commit

PetaPoco/Core/DateTime2Extentions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
```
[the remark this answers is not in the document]
roggmaeh: fixed in new commit

PetaPoco/Core/DateTime2Extentions.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace PetaPoco
+{
+    public static class DateTime2Extentions
+    {
+        /// <summary>
+        /// Converts an object to its <see cref="DateTime2"/> representation
+        /// </summary>
+        /// <param name="o">The object to be converted</param>
+        /// <returns></returns>
+        public static DateTime2 ToDateTime2(this object o) => new DateTime2((DateTime)o);
```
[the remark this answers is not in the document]
roggmaeh: dont like this, every country has its own format

PetaPoco/Attributes/ColumnAttribute.cs
```diff
@@ -35,6 +35,14 @@ public class ColumnAttribute : Attribute
         ///     json)"
         /// </summary>
         public string UpdateTemplate { get; set; }
+        /// <summary>
+        ///     True if AnsiString / VARCHAR column
+        /// </summary>
+        public bool IsAnsi { get; set; }
```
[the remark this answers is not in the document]
roggmaeh: fixed in new commit

PetaPoco/Database.cs
```diff
@@ -673,6 +689,19 @@ private void SetParameterProperties(IDbDataParameter p, object value, PropertyIn
                     // Thanks @DataChomp for pointing out the SQL Server indexing performance hit of using wrong string type on varchar
                     p.DbType = DbType.AnsiString;
                 }
+                else if (t == typeof(DateTime2))
+                {
+                    var asValue = (value as DateTime2).Value;
```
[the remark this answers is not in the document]
roggmaeh: fixed in new commit

PetaPoco/Database.cs
roggmaeh: ?? operator dont work with DateTime

PetaPoco/Database.cs
roggmaeh: I need this changes because my "old" DataBase Model has AnsiString/VARCHAR as PK's

PetaPoco/Core/DateTime2Extentions.cs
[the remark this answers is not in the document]
asherber: It doesn't need to be format specific. You could do something like:

```
public static DateTime2 (this string input) => new DateTime2(DateTime.Parse(input));
```

My point was that I think you only want to be able to pass in a `DateTime` or a string.

PetaPoco/Database.cs
asherber: Good point. Then `asValue` will never be null, so you don't need a null check. I think you can just do `p.Value = (value as DateTime2).Value`.

PetaPoco/Database.cs
asherber: I'm not sure what you mean by that. My point is that I don't like the idea of digging into `PropertyInfo` here to get the attributes. If this is going to be handled by attributes, then I think it might be better to have the attributes read as part of `PocoColumn` and then pass that in. Take a look at the way `ForceToUtc` is handled; I think something similar should work.

PetaPoco/Database.cs
roggmaeh: the source can be nullable DateTime object, null check has to be done

asherber: Oh, are you worried about the `value` argument being null? In that case, your line 694 would throw a null reference exception (trying to check `Value` property on a null object).

This should work:

```c#
p.Value = (value as DateTime2)?.Value ?? DBNull.Value;

// or

var asValue = (value as DateTime2)?.Value;
p.Value = asValue ?? DBNull.Value;
```

roggmaeh: Compiler Error CS0019 on both

asherber: Whoops, need to cast `DBNull.Value` to `object`.

```c#
p.Value = asValue ?? (object)DBNull.Value;
```

roggmaeh: fixed in new commit

PetaPoco/Core/DateTime2Extentions.cs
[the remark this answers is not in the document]
roggmaeh: added in new commit

PetaPoco/Database.cs
roggmaeh: added PocoColumn as param

PetaPoco/Core/DateTime2Extentions.cs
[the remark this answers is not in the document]
asherber: @roggmaeh Thanks, but I still think you want to change the method that takes an object so that it only takes a `DateTime`. Right now, you've got

```c#
public static DateTime2 ToDateTime2(this object o) => new DateTime2((DateTime)o);
```

If do `object foo = 4; foo.ToDateTime2();`, which is perfectly legal code, I get an `InvalidCastException`. I don't think there's a point to offering this method as an extension on all objects, when we know that it only works if the object is a `DateTime.`

PetaPoco/Database.cs
asherber: I see that you've added `PocoColumn` as a parameter, but I think you can now remove the `PropertyInfo` parameter -- it can be accessed as `pc.PropertyInfo` and so doesn't need to be passed in explicitly.

asherber:
@roggmaeh Thanks for all of your work on this, just a couple more small things to clean up!

But per my [original comment](https://github.com/CollaboratingPlatypus/PetaPoco/pull/630#pullrequestreview-840961143), I'm still not convinced that there's a practical need for this. That is, I'm not sure that adding this functionality will result in any real performance improvements. If you have any insight, please share it here.

If @pleb decides to go ahead with this, then we'll also need some tests to cover everything. But for now I would just hold on.

PetaPoco/Core/DateTime2Extentions.cs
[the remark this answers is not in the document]
roggmaeh: added in new commit

PetaPoco/Database.cs
roggmaeh: changed in new commt

roggmaeh:
@asherber in my case i have for both ForceToAnsiString and ForceToDateTime2/DateTime2 massive performance improvements.

ForceToAnsiString/AnsiString: my PK is a VARCHAR (i won't talk about it :cold_sweat: ), so on every update the SQL Server has a CONVERT_IMPLICIT warning and poor performance [https://blog.sqlauthority.com/2018/06/11/sql-server-how-to-fix-convert_implicit-warnings/](url)
UPDATE's with this fix from 7sec to 0.1sec

ForceToDateTime2/DateTime2 : CONVERT_IMPLICIT warning (not on all SQL Server Versions) when used in WHERE CLAUSE

asherber:
I've been reading up on this issue today. As best as I can tell, `CONVERT_IMPLICIT` only has a real performance effect when the column is in the `WHERE` clause, not when it's a value being inserted or updated. This is because it affects SQL Server's ability to use indexes to find the desired record(s).

In your case, with a `varchar` primary key, you're running into issues because `db.Update(poco)` generates a `SqlCommand` equivalent to `UPDATE table ... WHERE key_field = N'value'`, and that `nvarchar` doesn't play nice with the index on your `varchar` column. So I can see why being able to decorate a poco property to indicate that it should be `AnsiString` is useful.

I've also done some further reading about `datetime` and `datetime2`. In some quick testing, although there is a `CONVERT_IMPLICIT` going on for `SELECTs`, it doesn't appear to impact performance the same way that strings do, although I may have missed something. However, I do see the need for this on updated or inserted values, since `datetime` not only is less precise than `datetime2` but also rounds values to the nearest 3 ms. (Note that you can force the use of `datetime2` for inserted or updated values by setting `InsertTemplate` and `UpdateTemplate` on the poco property, but the syntactic sugar of `ForceToDateTime2` is useful.)

All of which is to say that I now understand better where all of this is coming from.

asherber:
@pleb When you have a chance to look at this, a few things:

1.  I suggested changing the names from `IsAnsi/IsDateTime2` to `ForceToAnsiString/ForceToDateTime2`, but now that I look at it some more, I'm not so sure (especially because `ForceToUtc` actually works in the other direction, controlling the FromDb mapping). I'd be okay with either `IsAnsiString` or just `AnsiString`.
2.  Does it feel like these want to be properties on the `Column` attribute, as they currently are, or separate attributes? Probably the former, but I'm not positive.
3.  The new `DateTime2` class should probably be modified so that `Value` is `DateTime?` instead of `DateTime`.
4.  Does the overall approach look right to you? What do you think about passing `PocoColumn` into `SetParameterProperties()` instead of `PropertyInfo`? 
5.  Still needs some tests around the new functionality.

@roggmaeh Thanks again for all of your work!

pleb:
Sorry, everyone. Happy to have a look, but it'll be a few more days till I get the chance

pleb:
Hey @roggmaeh. May I ask for what reason is it necessary to add specific handling for DateTime2?

asherber:
@pleb `datetime2` is (or at least can be) much more precise than `datetime`. When a C# `DateTime` or `string` is passed to SQL Server as `datetime`, it gets truncated and rounded -- even if it's being stored in a `datetime2` field.

Given this table:

```sql
create table Date_Time
(
  dt DateTime,
  dt2 DateTime2
)
```
Insert a pair of seemingly identical values, but pass one as `datetime` and one as `datetime2`:

```c#
Connection.Open();
var cmd = Connection.CreateCommand();
cmd.CommandType = CommandType.Text;
cmd.CommandText = "insert into Date_Time values (@0, @1)";

var p0 = new SqlParameter("@0", DbType.DateTime);
p0.Value = "2021-01-09 14:17:00.004";
cmd.Parameters.Add(p0);

var p1 = new SqlParameter("@1", DbType.DateTime2);
p1.Value = "2021-01-09 14:17:00.004";
cmd.Parameters.Add(p1);

cmd.ExecuteNonQuery();
```

Then pull them out (this is LinqPad):

```csharp
void Main()
{
    Connection.Open();
    using (var db = new Database(Connection))
    {
        db.Fetch<Date_Time>().Dump();
        //  DT = 14:17:00.0030000
        // DT2 = 14:17:00.0040000
    }
}

class Date_Time
{
    public DateTime DT { get; set; }
    public DateTime DT2 { get; set; }
    
    object ToDump() => new { DT = DT.ToString("HH:mm:ss.fffffff"), DT2 = DT2.ToString("HH:mm:ss.fffffff") };
}
```

I've never worked with `datetime2` in SQL Server myself, so I don't have a sense for how widespread this issue is. The type has been around since SQL Server 2008, and neither this library nor NPoco has added explicit support for it.

pleb:
@asherber thanks for the explanation. I knew of the loss when storing as DateTime verse DateTime2. I didn't know the `DbType` was involved.

1. Agree with AnsiString
2. Interesting... Yeah, I'm not 100% sure. My feeling is for separate attributes. I know there's `ForceToUtc`, but I think could be deprecated in favour of a replacement attribute too, as it seems a little out of place.
3. We should handle nulls out-of-the-box, but as you said a few tests would be good to ensure we're catering for the general usage patterns
4. I don't mind it. It seems to make for cleaner code
5. Agree

@roggmaeh Thanks for the PR. There's probably a bit of back and forth to go before it'll be accepted. Please, don't lose heart :) The reason is we need to be careful when making changes to the core API surface. It's hard to undo stuff in a popular lib, such as this, once you've shipped.

asherber:
@pleb My thought about attributes is that the `Column` attribute currently collects information about the various ways in which the property _interacts with_ the table column: name, insert template, update template, force utc. `AnsiString` and `DateTime2` seem to me to belong in this group as well. The `Ignore` and `ResultColumn` attributes are separate, but they have to do with _whether_ the column gets mapped, not _how_ the column gets mapped.

roggmaeh:
@pleb is there anything i can do. I need a release

pleb:
Sorry, @roggmaeh. It slipped by me. Let's roll with the changes you've made ;)

PR #642, opened 2022-05-26
Fix calls to GetCustomAttributes()

### LINKED ISSUE #640: Inherited attributes on properties are skipped
I discovered this today when looking into d-p-y/statically-typed-poco-queries#10

```c#
public class Parent
{
	[Column("Foo")]
	public virtual int ID { get; set; }
}

public class Child : Parent
{
	public override int ID { get; set; }
}

db.Query<Child>();
// EXPECTED: SELECT [Foo] FROM [Child]
// ACTUAL: SELECT [ID] FROM [Child]
```

The reason for this is that in `ColumnInfo`, when PP looks for attributes on a property, it does  `pi.GetCustomAttributes(typeof(xxxAttribute), true)`. As the Microsoft docs [indicate](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.getcustomattributes?view=net-6.0). the second parameter to `GetCustomAttributes()`, which indicates whether to look for inherited attributes, is ignored when you're looking at properties. What this really means is that it behaves as though the value were `false` and ignores inherited attributes. Oddly, if you omit the second parameter completely, the code follows a different path and _does_ return inherited attributes (which is the behavior we would prefer here).

This affects `ColumnAttribute`, `IgnoreAttribute`, and `ResultColumnAttribute`. I think it would also affect a `ValueConverterAttribute` in the parent class.

Note that in the (probably) more usual case of an inherited but not overridden property, everything is fine.

```c#
public class Parent
{
	[Column("Foo")]
	public virtual int ID { get; set; }
}

public class Child : Parent
{
}

db.Query<Child>();
// EXPECTED: SELECT [Foo] FROM [Child]
// ACTUAL: SELECT [Foo] FROM [Child]
```

I understand that this is an edge case -- an overridden property in a child class that doesn't also override the attribute. I've never come across it in real life. And fixing it would be a breaking change, since it would change the established public behavior. So if you decide you don't want to address it, that's fine -- I just wanted to bring it up for awareness.

asherber:
Fixes #640 

I verified that the tests (specifically Parent2/Child2) failed with the current code and pass after this change.

I wound up using `Attribute.GetCustomAttributes()` because the overload I wanted of `pi.GetCustomAttributes()` doesn't exist in net40.

### DIFF
--- a/PetaPoco.Tests.Unit/Core/ConventionMapperTests.cs
+++ b/PetaPoco.Tests.Unit/Core/ConventionMapperTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using PetaPoco.Tests.Unit.Models;
 using Shouldly;
 using Xunit;
 
@@ -269,6 +270,17 @@ namespace PetaPoco.Tests.Unit.Core
             public Guid Id { get; set; }
 
             public int Size { get; set; }
+        }
+
+        [Theory]
+        [InlineData(typeof(Child1))]
+        [InlineData(typeof(Child2))]
+        [InlineData(typeof(Child3))]
+        [InlineData(typeof(Child4))]
+        public void Converters_Should_Inherit(Type type)
+        {
+            var func = _mapper.GetToDbConverter(type.GetProperty("ID"));
+            func.ShouldNotBeNull();
         }
     }
 }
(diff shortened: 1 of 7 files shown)


asherber:
Note: This branch includes the `fix-unit-test` branch. It's probably better to merge that one first and then this one, so that the history is clear. Or just merge this one and discard the other.

Issue #644, opened 2022-06-10
IDbCommand is not guaranteed to be disposed 


maxisoft:
Hello,

I was browsing a code stacktrace to find memory leaks and I found out that there is a non disposed potential issue in [PetaPoco](https://github.com/CollaboratingPlatypus/PetaPoco).

Assuming that:
- `Database.CreateCommand()`  create a `System.Data.IDbCommand` which is *IDisposable*.
- The IDbCommand object is expected to be disposed by the caller.

Issue is that the code between `IDbCommand command = connection.CreateCommand()` and the `Database.CreateCommand()` return points is not guaranteed to **not throws**.

Example of such unsafe call :
- all the abstract property setter of a IDbCommand
- `ParametersHelper.ProcessQueryParams` may throws ArgumentOutOfRangeException

In such case the IDbCommand is not disposed (involving potential memory leak, stale sate, ...).

https://github.com/CollaboratingPlatypus/PetaPoco/blob/dd0ec718dd584dde45000f52c0df1213f32c9878/PetaPoco/Database.cs#L719-L740

pleb:
@maxisoft I think PR #645 addresses this

PR #645, opened 2022-06-11
Ensure db command is disposed

pleb:
Addresses #644


asherber:
(Technically, line 725 is not needed. `connection.CreateCommand()` automatically sets this property on the command. But there's no harm in leaving it.)

asherber (approved):

Issue #647, opened 2022-06-23
Accessing an Azure SQL database via access token?


Jepplar:
I am running an application in Azure which uses a managed identity, which means that I don't have a standard username/password combo in my connection string. Rather, the application uses a separate access token outside of the connection string to connect to the database.
How can I configure PetaPoco to use this access token?

Using the standard SqlClient, this example works as expected:
````
private int ExampleTest()
{
    string connectionString = "Server=xyz.database.windows.net,1433;Database=abc"; // does not configure a user/pw
    string query = $"...'";

    using SqlConnection connection = new();
    connection.AccessToken = this.GetToken(); // access token is set here
    connection.Open();
    SqlCommand command = new(query, connection);

    int count = (int)command.ExecuteScalar();
    return count;
}
`````

I was unable to figure our how to set the same using PetaPoco. I am using the SqlServerDatabaseProvider, but could not find an option inside it to configure an access token.
Is it only possible using a custom provider, and wrapping SqlClient myself, or can it be configured in the standard provider?

Many thanks for your time and input

**Comment by pleb** [comment]
@Jepplar yep, you can handle the connection yourself, and have PetaPoco do the heavy lifting.

    private int ExampleTest()
    {
        string connectionString = "Server=xyz.database.windows.net,1433;Database=abc"; // does not configure a user/pw
        string query = $"...'";

        using SqlConnection connection = new();
        connection.AccessToken = this.GetToken(); // access token is set here
        connection.Open();

        var petapoco = new PetaPoco(connection)
        return petapoco.Single<int>("select 1")
    }

That said, using a managed identity in azure shouldn't be a special case. 

💡Oh, I see `conn.AccessToken = token.Token;`

Hmm. I guess it would be nice for the DbProvider to be able to set this and/or have a pre-con open event/hook

**Comment by Jepplar** [comment]
That sounds great 👍 I will wait for the PR to be merged and try it the new way!

**Comment by asherber** [comment]
You could also do:

```c#
using (var db = new Database(...))
{
	((SqlConnection)db.Connection).AccessToken = "myToken";
}
```

pleb:
> You could also do:
> 
> ```cs
> using (var db = new Database(...))
> {
> 	((SqlConnection)db.Connection).AccessToken = "myToken";
> }
> ```

I realised this method only works if the connection is passed via the constructor, as `Connection` is `null` until a connection is opened

pleb:
[...]

PR #648, opened 2022-06-23
Adds a hook to allow connection modification before open is called

pleb:
Helps solve #647


asherber:
I think `OnConnectionOpening` etc. is probably more idiomatic than `OnConnectionPreOpened`.

PR #663, opened 2023-01-18
Fix ReferenceAssemblies dependency

asherber:
This appears to be my fault -- when I was doing some work to add a .net standard 2.1 target, it looks like I screwed up the dependency on ReferenceAssemblies. It should only be a build-time dependency, but it was getting linked in to the NuGet package. No harm, just some files dragged in that shouldn't be.

### DIFF
--- a/PetaPoco/PetaPoco.csproj
+++ b/PetaPoco/PetaPoco.csproj
@@ -21,8 +21,8 @@
     <Reference Include="System.Configuration" />
   </ItemGroup>
  
-  <ItemGroup>
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies.net40" Version="1.0.2">
+  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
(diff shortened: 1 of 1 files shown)


pleb:
I've been working on this [one](https://github.com/CollaboratingPlatypus/PetaPoco/compare/development...feature/petapoco-v7) on and off. It would bring the lowest supported standard to 2.1. It also moves the CI/CD to github actions and runs all unit and integration tests.

asherber:
IMO, the only reason to include netstandard2.1 is to make use of a feature that's particular to that spec -- for example, async transactions in PP. But making it the _only_ supported framework means that nobody on .NET Framework [can use the library](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1), and nobody with .NET Core apps \< 3.0. 

So for best usability I think it's advisable to keep netstandard2.0 in the mix. It's also nice to keep the net40 and net45 targets because (a) anyone using framework \< 4.6.1 can't use a netstandard2.0 library, and (b) using a netstandard2.0 library from framework > 4.6.1 means you're dragging in a bunch of other netstandard libs that you really don't need. You might be surprised at how many active applications still target older .NET versions; since an app can target any 4.x and still work on the 4.8 runtime included with current Windows, there's no particular reason to upgrade unless (as above) you want to use newer features.

From a testing perspective, this means that tests ideally should run on a Windows box so that they can target some Framework version along with multiple .NET (Core) versions.

pleb:
I figured version 6 would still be an option for < 2.1. However, maintaining the two would be a pain. I guess as you pointed out, the sweet spot is currently 2.0. I'll keep the testing environment OS in mind. I believe it should be fine to use Windows

PR #670, opened 2023-05-13
add null check for property `Set` method

### LINKED ISSUE #664: Improve error handling when unsettable properties are being used
Imagine a POCO class like this:

```c#
public class TransactionCategorizationRule
{
    public int Id { get; }
}
```

...and then try to map to this class using like this:

```c#
var transactionCategorizationRules = db.Query<TransactionCategorizationRule>("SELECT * FROM transaction_categorization_rule")
    .ToImmutableList();
```

This will yield the following exception:

```
Unhandled Exception: System.ArgumentNullException: Value cannot be null. (Parameter 'meth')
   at System.Reflection.Emit.DynamicILGenerator.Emit(OpCode opcode, MethodInfo meth)
   at PetaPoco.Core.PocoData.<>c__DisplayClass29_0.<GetFactory>b__0()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at PetaPoco.Database.ExecuteReader[T](CommandType commandType, String sql, Object[] args)+MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at System.Collections.Immutable.ImmutableExtensions.FallbackWrapper`1.get_Count()
   at System.Collections.Immutable.ImmutableList`1.CreateRange(IEnumerable`1 items)
   at System.Collections.Immutable.ImmutableList`1.AddRange(IEnumerable`1 items)
   at System.Collections.Immutable.ImmutableList.ToImmutableList[TSource](IEnumerable`1 source)
   at Poc.Hemekonomi.ConsoleApp.Program.CategorizeTransactions() in /home/per/git/poc.hemekonomi/Poc.Hemekonomi.ConsoleApp/Program.cs:line 131
   at Poc.Hemekonomi.ConsoleApp.Program.<>c__DisplayClass2_0.<Main>b__2() in /home/per/git/poc.hemekonomi/Poc.Hemekonomi.ConsoleApp/Program.cs:line 33
```

Of course, the non-settable property is the "root cause" here, but it would be nice if we would fail more gracefully in scenarios like this. (https://github.com/DapperLib/Dapper/issues/912 was helpful in debugging this, btw)

Thanks for a nice little tool! :slightly_smiling_face: :+1:

Ste1io:
- [x] Add tests

fixes #664

### DIFF
--- a/PetaPoco.Tests.Integration/Databases/BaseQueryTests.cs
+++ b/PetaPoco.Tests.Integration/Databases/BaseQueryTests.cs
@@ -224,6 +224,21 @@ namespace PetaPoco.Tests.Integration.Databases
         }
 
         [Fact]
+        public void Query_ForPocoWithPropertyMissingSetMethod_ShouldThrow()
+        {
+            AddPeople(1, 1);
+
+            var pdPerson = PocoData.ForType(typeof(Person), DB.DefaultMapper);
+
+            var dbPeople = DB.Provider.EscapeTableName(pdPerson.TableInfo.TableName);
+            var pFullName = DB.Provider.EscapeSqlIdentifier(pdPerson.Columns.Values.Single(c => c.PropertyInfo.Name == nameof(Person.Name)).ColumnName);
+
+            var sql = new Sql($"SELECT {pFullName} FROM {dbPeople}");
+
+            Should.Throw<InvalidOperationException>(() => DB.Query<ReadOnlyPoco>(sql).ToList());
+        }
+
+        [Fact]
         public void Query_ForMultiPocoWithWildcard_ShouldReturnValidPocoCollectionAfterColumnAdded()
         {
             AddOrders(1);
@@ -246,6 +261,28 @@ namespace PetaPoco.Tests.Integration.Databases
 
             results = DB.Query<Order, Person>(testQuery).ToList();
             results.ShouldNotBeEmpty();
+        }
+
+        [Fact]
+        public void Query_ForMultiPocoWithPropertyMissingSetMethod_ShouldThrow()
+        {
+            AddOrders(1);
+
+            var pdOrder = PocoData.ForType(typeof(Order), DB.DefaultMapper);
+            var pdPerson = PocoData.ForType(typeof(Person), DB.DefaultMapper);
+
+            var dbOrders = DB.Provider.EscapeTableName(pdOrder.TableInfo.TableName);
+            var dbPeople = DB.Provider.EscapeTableName(pdPerson.TableInfo.TableName);
+
+            var oId = DB.Provider.EscapeSqlIdentifier(pdOrder.Columns.Values.Single(c => c.PropertyInfo.Name == nameof(Order.Id)).ColumnName);
+            var pId = DB.Provider.EscapeSqlIdentifier(pdPerson.Columns.Values.Single(c => c.PropertyInfo.Name == nameof(Person.Id)).ColumnName);
+
+            var oPersonId = DB.Provider.EscapeSqlIdentifier(pdOrder.Columns.Values.Single(c => c.PropertyInfo.Name == nameof(Order.PersonId)).ColumnName);
+            var pFullName = DB.Provider.EscapeSqlIdentifier(pdPerson.Columns.Values.Single(c => c.PropertyInfo.Name == nameof(Person.Name)).ColumnName);
+
+            var sql = new Sql($"SELECT o.{oId}, p.* FROM {dbOrders} o JOIN {dbPeople} p ON o.{oPersonId} = p.{pId}");
+
+            Should.Throw<InvalidOperationException>(() => DB.Fetch<ReadOnlyMultiPoco, Person>(sql).ToList());
         }
 
         [Fact]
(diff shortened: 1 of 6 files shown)


pleb:
Any chance of adding a test to confirm? See https://github.com/CollaboratingPlatypus/PetaPoco/blob/development/PetaPoco.Tests.Integration/Databases/MSSQL/MssqlQueryTests.cs for how existing bugs have been tested

Ste1io:
@pleb Thanks for the reference;; you read my mind, as I had planned to ask for it once one of the maintainers responded. I had a chance to clone my fork to my pc earlier today with the intention of adding a test, but wasn't sure where it should go exactly. Been using the library since the early "single-file" days of v4, but today was my first time digging into the tests projects, so a good bit of the code base as it stands today feels more Déjà vu than familiar, lol.

I'll add the test(s) push them to this branch this weekend.

Ste1io:
@pleb This may be better placed in BaseQueryTests.cs as it's not specific to any single db engine, didn't want to alter all the scripts until I get a confirmation and test is considered complete.

pleb:
> @pleb This may be better placed in BaseQueryTests.cs as it's not specific to any single db engine. I didn't want to alter all the scripts until I get a confirmation and test is considered complete, however.

Ah, good point. There's no technical issue with these testing living where they are, as this test covers all providers as it's code a single provider cannot influence behaviour over. That said, future maintainers might assume the bug was with this MySQL, so moving higher makes sense.

Re coverage, I think what you have there is fine 👍 

PS. Thanks for contributing :)

Ste1io:
> That said, future maintainers might assume the bug was with this MySQL, so moving higher makes sense.

I'll take a look this evening and get this sorted.

On another note, instead of just outright throwing if `GetSetMethod` fails, any reason not to first attempt to set the backing field (if present)? That would allow full bodied properties to still work, even without a `Set` accessor:

```csharp
public class MyClass
{
	private int _id; // <- set the field directly
	public int Id => _id;
}
```

asherber:
If a property doesn't have a public setter, then the class author didn't want it to be settable.

Ste1io:
@pleb I opted to leave the tests where they are; there is a comment noting that they are not specific to any one db engine to prevent confusion for future contributors who may not review the file blame... Moving them to the base class with the four new structs meant either breaking consistency and defining the structs inline with the test cases, or adding an additional file to the base test class - neither of which seemed appropriate to me unless you feel strongly about it.

@asherber I agree that the lack of a setter does imply intent for users of an API, but thought it worth mentioning just in case. As a PetaPoco user, I view the library in my projects as an extension of my (hidden) implementation, and prior to dotnet adding support for private Property setters (which are assigned to in this case), replacing the public setter with a backing field that is mutable both by my class and (by extension) PetaPoco would have made sense to me.

Regardless, avoiding ambiguity is ultimately far more important that adding a small "feature" that could be considered a breaking change, so I digress. :)

Thanks again for continuing to maintain such an amazing library - it was the first ORM I ever used, and after using it in countless projects, it's always done right by me, and I'm grateful to finally be able to give back something, even if just a little.

asherber:
I defer to @pleb, but I'd rather see this moved to `BaseQueryTests`, even it means adding a new model class to go with it. I don't think you actually need to create a new table in each db; since you're just looking at the `Id` property, you can use any of the existing tables with an `Id` column.

Ste1io:
Valid point, I would agree. Iirc, a handful of the MSSQL tests would appear to fall under the same category. Should those be hoisted up as well, provided that is the case?

pleb:
> Valid point, I would agree. Iirc, a handful of the MSSQL tests would appear to fall under the same category. Should those be hoisted up as well, provided that is the case?

No need to do it in this PR. If you have time, it would be great to refactor these in another PR

Ste1io:
> No need to do it in this PR. If you have time, it would be great to refactor these in another PR

Sounds good, will make a new feature branch for the other tests this week or over the weekend. Regarding my tests from this bugfix specifically, I'll update their location as discussed and commit to this branch tonight or tomorrow so they aren't holding up the PR any longer.

asherber:
I'd like to suggest that you squash all your commits when you're done, to help make things less busy.

asherber:
@ste1io Why merge `development` in? It's just an extra commit in your branch, and it prevents you from squashing. I would suggest resetting your branch to the previous commit (looks like `778d`) and then squashing your commits down to `de1b` and force pushing. That way you'll have just a single commit on this PR, making for a cleaner merge into `development`.

Ste1io:
@asherber already on it, that was an error on my part not paying attention to which branch I was pulling. Good catch.

Ste1io:
We should be all set with this PR at this point @pleb.

By the way, there was an oversite with the code coverage from my original multipoco test which I would have never caught if I hadn't changed the tests to use the current db tables like @asherber suggested, so refactoring and moving the tests as you both recommended definitely paid off. :)

PR #678, opened 2023-06-07
Link fixes, minor textual changes to repo health files

Ste1io:
I'm opening this PR a bit earlier than I had planned, but it addresses several issues with broken links in the README, including replacing the now defunct CodeHorror link - the original domain appears to have expired and changed hands, and is now part of what appears to be a malware redirect chain. RIP.

This PR does rename `contributors.md` to all uppercase, though only for consistency with the file casing conventions used by the other health files. If there is preference with leaving it unchanged, please say. All content changes for that file are in a separate commit for easier diffing.


pleb:
Nice work, @Ste1io ❤️

Ste1io:
Cheers @pleb, happy to help.

pleb:
I've updated the license. I originally changed it from Apache 2 to MIT but reverted to an updated Apache 2. Personally, I like the MIT license because it's easier to read/understand, but I figure a change might cause downstream pain for users who are compliant with the Apache 2

PR #681, opened 2023-06-17
Add GenerateDocumentationFile element to csproj

### LINKED ISSUE #680: XML Documentation missing in Nuget package
I believe it's due to missing the "GenerateDocumentationFile" element in the `.csproj` file.

Ste1io:
This should fix #680, based on [this StackOverflow answer](https://stackoverflow.com/a/48664173).

### DIFF
--- a/PetaPoco/Core/IProvider.cs
+++ b/PetaPoco/Core/IProvider.cs
@@ -1,4 +1,4 @@
-using System.Data;
+﻿using System.Data;
 using System.Data.Common;
 #if ASYNC
 using System.Threading;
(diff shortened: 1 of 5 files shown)


Ste1io:
Account to [the documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/#create-xml-documentation-output):

I will run a build from my PC this morning and add any missing xml doc comments, in case you want to hold off till then to merge this.

Ste1io:
Warnings:

`1591` - 195 (XML comment missing)
`1574` - ~~3 (XML comment with unresolvable cref attribute)~~ (fixed)
`1572,1573` - ~~2 (XML comment with missing or mismatched param tags)~~ (fixed)

I just added a commit which fixes the last 4 types mentioned above. The missing comments I'll start tackling today, hopefully have it ready before the weekend is over but will see how it goes. Should I just do that as a separate branch with it's own PR so this one isn't held up? Having *most* doc comments available for nuget users is far better than having *none*, imo. :)

PR #682, opened 2023-06-23
XML doc comments coverage (edits & additions)

Ste1io:
A couple notes regarding the changes on this branch:

- All multi-line comment sections (`<section>...</section>`, `<remarks>...</remarks>`, etc) have been combined into single lines.
- Indented comments had the preceeding spaces removed (`///     foobar` => `/// foobar`).
- The order of some xml doc comment blocks were rearranged based on the following ordering rules:
  ```xml
  <!-- methods -->
  <section/>
  <remarks/>
  <typeparam/>
  <param/>
  <returns/>
  <exception/>

  <!-- properties -->
  <section/>
  <remarks/>
  <value/> <!-- used instead of "returns" tag -->
  ```

The above changes were to enable me to more easily parse through the existing comments to identify and fix any spelling errors or needed modifications. Also, it should make reviewing the diffs less cumbersome. Not to mention editing or writing comments while maintaining the appropriate line breaks is a pain in the @$$. :) Easier to just reapply the formatting when everything is ready for merging.

While my intention was to just add the missing doc comments causing the compiler warnings, it made more sense for me to do a full thorough review while I was at it. Ensuring consistency across the board will make implementing a static documentation site for GitHub pages (using docfx or something similar) that's generated from the code *sooo* much easier, if we choose to do that (imo, it would be a good idea). Most importantly, though, it's been very beneficial for me personally catching up with the changes to the library since the single-file days, which is helping shape out the direction for some of the new documentation tremendously.

Apologies in advance for such a huge amount of code for you guys to look over. I plan on squashing the commits when it's ready for merging, but leaving them as is in the meantime so that reviewing related changes between two consecutive commits is possible. Most of the [changes in the first commit](https://github.com/CollaboratingPlatypus/PetaPoco/pull/682/commits/5c00002e12186ffdb14071c8c42c475672a1829d) will evaporate with the final commit, but actual [changes made to the existing comments](https://github.com/CollaboratingPlatypus/PetaPoco/pull/682/commits/39b9133581fdab2000378bb28dc777cc62c90fef) I felt should be easily reviewed without formatting changes or additions crowding up the diffs.


.editorconfig
Ste1io, line 1: I generated the .editorconfig from the current source's styles. My personally settings, using tabs for indentations, kept sneaking in to my commits on accident, so here's to no more unnecessary white-space-only file diffs. xd

PetaPoco/EventArgs.cs
Ste1io, line 1: I just noticed that the `DbConnectionEventArgs.IDbConnection Connection` property has both a getter and setter unlike the others. Is this property meant to be mutable after construction?

Ste1io:
This puts us at 80 remaining.

PetaPoco/OracleProvider.cs
Ste1io, line 1: This file probably belongs in the "Providers" directory with the others. Any objections to me moving it?

asherber: I guess I don't mind moving it physically, but the namespace can't be changed.

PetaPoco/EventArgs.cs
asherber: Well, I added those EventArgs classes, along with the events in `Database`. The goal was to provide an easier way to customer an object -- add a handler to `ConnectionOpened` instead of having to define a descendant class and override `OnConnectionOpened`. 

You'll note that the (existing) docs for `OnConnectionOpened` allow for returning a proxy or replacement connection, so I guess I followed that pattern when creating the corresponding event/args. Is it a good idea? Is anybody using it that way? I can't say.

PetaPoco/OracleProvider.cs
Ste1io: If you're not too keen on moving it, then no real reason to. I just wasn't sure if had been overlooked or not, since it hasn't made it into the test bed yet.

asherber: Oh, actually, it's not a `DatabaseProvider` -- it's an implementation of `DbProviderFactory`, and it's been around since the Brad Robinson days. Almost certainly no one is using it anymore.

PetaPoco/Utilities/PagingHelper.cs
Ste1io, line 1: Putting this here just to ensure it doesn't get missed; I converted the public fields to properties with get accessors; assigning a default value so it should perform essentially the same as the fields. If they were left as fields for a reason before, let me know, but making them readonly properties seemed like the safer way of exposing them imho.

There's one other place with the same issue also, with the public fields in the `PocoColumn` class (though in this case they'd be full properties with get/set methods):

https://github.com/CollaboratingPlatypus/PetaPoco/blob/a48db2e1c01fbc882733ef8412fb456aab023760/PetaPoco/Core/PocoColumn.cs#L8-L16

Should these two changes be done through a different PR, or is including them in this one an issue?

PetaPoco/OracleProvider.cs
pleb: When I took over from Brad, I didn't touch the Oracle code because there wasn't an easy way of testing it. I left it as is. I'm not interested in maintaining Oracle support myself, but I left it there in case people still use it. If they do and they have an issue, they're more than welcome to contribute a fix :)

PetaPoco.Tests.Integration/Documentation/Delete.cs
```diff
@@ -20,7 +20,10 @@ public Deletes()
         [Fact]
         public void DeleteByPoco()
         {
-            // Create the person
+            // Clear out any people and reset the ID sequence counter
+            DB.Execute("TRUNCATE TABLE [People]");
```
pleb, line 24: I thought each test had a fresh DB instance?

pleb: They do. If you look at the build scripts, they should all reset the DB and then build it back up again. Ref https://github.com/CollaboratingPlatypus/PetaPoco/blob/development/PetaPoco.Tests.Integration/Scripts/MySqlBuildDatabase.sql

PetaPoco.Tests.Integration/Documentation/Delete.cs
```diff
@@ -118,22 +121,25 @@ public void DeleteCustomSqlWhere()
         [Fact]
         public void DeleteAdvanced()
         {
-            // Create the person
+            // Clear out any people and reset the ID sequence counter
+            DB.Execute("TRUNCATE TABLE [People]");
```
pleb, line 125: https://github.com/CollaboratingPlatypus/PetaPoco/pull/682/files#r1241341082

PetaPoco.Tests.Integration/Documentation/Inserts.cs
```diff
@@ -75,6 +75,9 @@ public void InsertAutoIncrement()
         [Fact]
         public void InsertToDifferentTable()
         {
+            // Clear out any people and reset the ID sequence counter
+            DB.Execute("TRUNCATE TABLE [People]");
```
pleb, line 79: https://github.com/CollaboratingPlatypus/PetaPoco/pull/682/files#r1241341082

pleb:
@Ste1io thanks for fixing this up :)

PetaPoco.Tests.Integration/Documentation/Delete.cs
Ste1io: My reasoning for adding that was due to the fact that 1) the documentation tests are used (in the docs) independent of the test suite, and 2) some of the tests *did* include that reset. Since the assertions require an empty table each run to pass anyways, and as they're presented outside the context of the actual test project, the added consistency seemed justifiable.

PetaPoco/OracleProvider.cs
Ste1io: Well, it's cleaned up a bit at least (as far as green text is concerned); beyond that, not going to bother with it.

Ste1io:
Morning gents, wrapping up the remaining xml doc comments for all things async before the weekend. Do you want the leading spaces for summary and remarks sections added back in, or left as is? Also, what's the magic number for the column breaks?

pleb:
@Ste1io I don't mind regarding leading space or not, so long as it's consistent. Regarding doc comments length, pick something that works for you but isn't too constrained for everyone... maybe something between 80 - 160 etc

Ste1io:
@pleb will do. Plan on wrapping this PR up over the weekend or early next week. What do you want me to do about the changes to the doc tests?

PetaPoco.Tests.Integration/Documentation/Delete.cs
pleb: Oh, I see. I am fairly sure the docs tests don't need to `truncate`. Could you try removing it? The idea is that all integration tests start with a fresh database, and how that's done is left to the database script. This is important because we don't have/shouldn't ever need to place vendor-specific SQL in the base tests.

Ste1io: Having consistency between the documentation tests was my main concern here, but I do agree with you they should be removed, and will include it in the next commit. (Note the discrepency between [`DeleteByPoco()` and `DeleteByPrimaryKey()`](https://github.com/CollaboratingPlatypus/PetaPoco/wiki/Deleting#examples) in the wiki).

Since the `TRUNCATE` commands were in *some* but not all of the doc tests for Insert/Delete/Update/Save previously, as in the link above, you want those removed also I presume? I'm pretty sure they are only there as artifacts from the early days prior to integrating XUnit.

pleb: Sorry. Yes please, @Ste1io. 

> I'm pretty sure they are only there as artifacts from the early days prior to integrating XUnit.

I wish that were the case, but it's probably more likely it was, ehm, me. Early on, after I took over from Brad, there was little in the way of code reviews :)

Ste1io:
Aside from some minor fix-ups and any additional changes following code review, xml doc comments are finished. So as not to contribute to the loss of your sanity, I batched all relocated code into two commits, so the actual changes can be easily diffed without making your eyes bleed. Expect these commits to get pushed to remote shortly.

Ste1io:
https://github.com/CollaboratingPlatypus/PetaPoco/pull/682/commits/e294f5bb06812d28ea90c6f96e59f44cec010fc3 is the best commit to compare diffs for additional changes after this point, as everything is a line-for-line match.

Ste1io (commented):
This is the best commit to compare diffs for additional changes after this point, as everything is a line-for-line match.

PetaPoco/IAsyncReader.cs
```diff
@@ -17,7 +17,9 @@ public interface IAsyncReader<out T> : IDisposable
         /// <summary>
         /// Asynchronously reads the next row from the data source.
         /// </summary>
-        /// <returns>A task that represents the asynchronous read operation. The task result contains <see langword="true"/> if more records exist; otherwise, <see langword="false"/>.</returns>
+        /// <returns>A task that represents the asynchronous operation. The task result contains <see langword="true"/> if more records exist, otherwise <see langword="false"/>.</returns>
         Task<bool> ReadAsync();
+
+        // TODO: Missing: `Task<bool> ReadAsync(CancellationToken cancellationToken);`
```
Ste1io, line 23: `IAsyncReader.ReadAsync()` should either accept a `CancellationToken` parameter, or enforce an overload with that parameter. The implementation for `AsyncReader` at https://github.com/CollaboratingPlatypus/PetaPoco/blob/5cd6bdbf208ede225dce461d44666bf6d9240092/PetaPoco/Utilities/AsyncReader.cs#L46 would need updated accordingly also.

Ste1io:
I've got a dozen or so issues that surfaced while doing the doc comments. They all **need** to be addressed, but they aren't related to my changes. I started adding them in this thread, but as some of them are breaking changes (changes with interface contracts, as one example), I'd rather avoid holding up this PR until a major release because preexisting issues are associated with it.

Any preference with how I go about them? Bulk them up in a couple of issues, or an issue for each? They're all fairly easy to fix, and I'll be happy to front the grunt work with fixing them (I have a local branch with many of the fixes already implemented, as it stands), but pushing to a release build would require a major version bump.

For context, here's a portion of the todo comment dump so you know what I'm referring to (objective observations only; opinionated ideas are omitted for another discussion on another day :)):

```
21 results - 7 files

PetaPoco\Database.cs:
   506:         // TODO: [BREAKING] Not implemented: `AbortTransactionAsync(CancellationToken)`
   527:         // TODO: [BREAKING] Not implemented: `CompleteTransactionAsync(CancellationToken)`
   572:         // TODO: [BREAKING] Not implemented: `CleanupTransactionAsync(CancellationToken)`
   579:         // TODO: [BREAKING] Not in IDatabase interface: `OnException(Exception)`, to allow GridReader's DI impl to accept IDatabase (IAsyncReader is able to correctly use the interface for it's IOC because it doesn't require the OnException))
   895:         // TODO: [BUGFIX] QueryAsync(CancellationToken, string, object[]) takes a caller-provided CancellationToken, but uses a default empty token for wrapped call
   949:         // TODO: [BUGFIX] QueryAsync(Action<T>, CancellationToken, string, object[]) takes a caller-provided CancellationToken, but uses a default empty token for wrapped call
  1340:         // TODO: [BUGFIX] FetchAsync(CancellationToken, string, object[]) takes a caller-provided CancellationToken, but uses a default empty token for wrapped call
  1622:         // TODO: [BUGFIX] SkipTakeAsync(long, long, Sql) should forward call to an overload that receives a CancellationToken
  3220:         // TODO: [BREAKING] method `CommandHelperAsync` takes DBCommand type; inconsistent with synchronous version `CommandHelper` which correctly uses IDbCommand type

PetaPoco\IAsyncReader.cs:
  23:         // TODO: [BREAKING] Missing: `Task<bool> ReadAsync(CancellationToken cancellationToken);`

PetaPoco\IConnection.cs:
  42:         // TODO: [BREAKING] Missing overload in IConnection interface: Task OpenSharedConnectionAsync(CancellationToken);

PetaPoco\IDatabase.cs:
  105:         // TODO: [BREAKING] Not in interface: `AbortTransactionAsync()`
  106:         // TODO: [BREAKING] Not in interface: `AbortTransactionAsync(CancellationToken()`
  115:         // TODO: [BREAKING] Not in interface: `CompleteTransactionAsync()`
  116:         // TODO: [BREAKING] Not in interface: `CompleteTransactionAsync(CancellationToken()`
  141:         // TODO: [BREAKING] Not in interface: `event EventHandler<DbConnectionEventArgs> ConnectionOpening`

PetaPoco\IQuery.cs:
  472:         // TODO: [BREAKING] Missing overload: `bool IQuery.Exists<T>(Sql sql);`

PetaPoco\IQueryAsync.cs:
  478:         // TODO: [BREAKING] Missing overload: `IQueryAsync.ExistsAsync<T>(Sql sql);`
  495:         // TODO: [BREAKING] Missing overload: `IQueryAsync.ExistsAsync<T>(CancellationToken cancellationToken, Sql sql);`

PetaPoco\Utilities\AsyncReader.cs:
  62:         // TODO: [BREAKING] Add `ReadAsync(CancellationToken)` overload
```

There's also a heap of "not broken but should be discussed" notes that would all fit into at least a partial refactor (v7?), which I've omitted here, as they should probably best be proposed with a discussion so everything is somewhat coherent. :)

pleb (approved):

Ste1io:
After @asherber is finished looking over anything and any changes following are sorted, I'll update this branch with the text wrap and we should be good to. You want me to squash down these commits on that final commit, or leave them as is and let you handle that how you want when you merge in the PR?

pleb:
All good, @Ste1io. If the tests are passing, I say it's good to go

Ste1io:
> All good, @Ste1io. If the tests are passing, I say it's good to go

All Sqlite tests pass. I need to get my docker setup sorted before I can test the others, but will advise as soon as that's done.

Ste1io:
Finally got around to restarting my pc (a rare occurrence) and reconfiguring virtualization in BIOS again since replacing my mobo and processor earlier this year, so I'm good to go with docker once more.

Which brings me to my question... How long has it been since the full test suite has been run for all dbms?

Ste1io:
All tests pass. Ready to merge @pleb 

1696104597-PetaPoco_-_Microsoft_Visual_Studio

Ste1io:
42crbn

Issue #675, opened 2023-05-25
When inserting, name of autoincrement column with expression is not escaped


iadaz:
I will check this myself and provide a PR, but I just wanted to get this down as a reminder until I get a chance to look at it properly. 

In [Database.cs](https://github.com/CollaboratingPlatypus/PetaPoco/blob/development/PetaPoco/Database.cs) / PrepareExecuteInsert() it optionally adds the autoincrement column name to the statement if there's an expression for it:

```
// Don't insert the primary key (except under oracle where we need bring in the next sequence value)
if (autoIncrement && primaryKeyName != null && string.Compare(i.Key, primaryKeyName, StringComparison.OrdinalIgnoreCase) == 0)
{
    // Setup auto increment expression
    var autoIncExpression = _provider.GetAutoIncrementExpression(pd.TableInfo);
    if (autoIncExpression != null)
    {
        names.Add(i.Key);
        values.Add(autoIncExpression);
    }

    continue;
}

names.Add(_provider.EscapeSqlIdentifier(i.Key));
values.Add(string.Format(i.Value.InsertTemplate ?? "{0}{1}", _paramPrefix, index++));
AddParam(cmd, i.Value.GetValue(poco), i.Value);
```

I think this should add `_provider.EscapeSqlIdentifier(i.Key)` as the section below for regular columns does.

Ste1io:
Possibly related to #603

iadaz:
I have made the change outlined above and it looks OK. I don't think I can provide a unit test for this though. My use case isn't Oracle, I am using a custom provider for SAP Hana which has similar syntax when using a sequence as a key. 

Are you happy to get a PR on this basis?

iadaz:
Just for the benefit of anyone else trying to create a custom provider for SAP Hana and comes across this, here's what I learned about Hana keys, both auto-increment and using sequences.

**Using a regular autoincrement key:**

There's no `RETURNING ... INTO` support on Hana. You can get the new key back with `select current_identity_value() FROM DUMMY` (but be aware of threading issues.)

This can't be done in one command execution because you can't execute multiple SQL statements (separated by a semi-colon) without wrapping it all in a BEGIN/END. But wrapping it makes it a Hana SQLScript block (instead of a prepared statement) and this then means the parameters are lost!

So your `ExecuteInsert` overload has to run the command query and then get the key back in a separate select.

**When the key comes from a sequence:**

[...]

PR #683, opened 2023-07-14
675 when inserting name of autoincrement column with expression is not escaped

iadaz:
As per https://github.com/CollaboratingPlatypus/PetaPoco/issues/675, autoincrement column names should be escaped.


Ste1io:
Deleted this branch now that it's been merged. Thank you for your contribution, @iadaz

PR #684, opened 2023-07-18
fix: ci/cd workflow build error

Ste1io:
~~Untested, but this should resolve~~ fixes the nuget source error causing appveyor builds to fail on development branch.

### DIFF
--- a/.github/workflow/pr.yml
+++ b/.github/workflow/pr.yml
@@ -2,6 +2,8 @@ name: PR Build and Test
 
 on:
   pull_request:
+    branches: [ development ]
+  push:
     branches: [ development, master ]
 
 jobs:
--- a/.github/workflow/pr.yml
+++ b/.github/workflow/pr.yml
@@ -2,7 +2,7 @@ name: PR Build and Test
 
 on:
   pull_request:
-    branches: [ development ]
+    branches: [ development, master ]
 
 jobs:
   runner-job:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,10 +17,11 @@ build_script:
 
     if ($env:APPVEYOR_REPO_BRANCH -ieq "development") {
         $packVersion += "-beta"
-        dotnet pack -c "Release" -p:Version=$packVersion
     }
-    elseif ($env:APPVEYOR_REPO_BRANCH -ieq "master") {
-        dotnet pack -c "Release" -p:Version=$packVersion
-        dotnet nuget push ".\PetaPoco\bin\Release\*.nupkg" -k "$($env:NUGET_API_KEY)" -s "https://www.nuget.org/api/v2/package"
+    
+    dotnet pack -c "Release" -p:Version=$packVersion
+    
+    if ($env:APPVEYOR_REPO_BRANCH -ieq "master") {
+        dotnet nuget push "./PetaPoco/bin/Release/*.nupkg" -k "$($env:NUGET_API_KEY)" -s "https://www.nuget.org/api/v2/package"
     }
 test: off
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ version: 6.0.{build}
 branches:
   only:
   - master
-  - development  
+  - development
 skip_tags: true
 max_jobs: 1
 image: Visual Studio 2022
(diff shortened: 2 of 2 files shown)


.github/workflow/pr.yml
```diff
@@ -2,7 +2,7 @@ name: PR Build and Test
 
 on:
   pull_request:
-    branches: [ development ]
+    branches: [ development, master ]
```
Ste1io, line 5: On second thought, reverting this change and using a separate workflow for master branch that's triggered on push (vs pull_request) would probably be more appropriate.

pleb: @Ste1io go for it. 

I started working on moving this over to Github actions, along with dropping some old stuff. Ref https://github.com/CollaboratingPlatypus/PetaPoco/tree/feature/petapoco-v7/.github/workflows

If you ever feel the need to do some devops, feel free to use add github action support. steal anything you want from my v7 branch 😉

Ste1io: Sounds good. DevOps is an enticing diversion from comments... 😂 Will update this branch once the other PR it's out of draft status, or possibly sooner from a worktree depending on my sanity level xd.

Ste1io: Was the plan for versioning moving forward with v7?

pleb: The plan for v7 was to drop old .net support, which means a path to remove

- Pre-processor #if !NETSTANDARD
- Pre-processor #if ASYNC

And to use Github actions and drop appveyor

pleb: LOL @ my commit comments. I assume any devops person does the same 👀

Ste1io: Regarding versioning specifically, though? Since the build number is currently handled with appveyor.

And yes, we all have those moments lmao.

pleb: Oh, sorry. I misunderstood.

Re the version, I still needed to solve this part of the puzzle. I assume the v7 was also for this, so there was a clean break between appveyor and github

Ste1io: I changed the triggers for the branches, so things should run without any change from the current CD, and CI no longer bugs out. Keep things in a stable state for the rest of this major version. Since you've already started the new workflow on v7's branch, I'm thinking best to pick up there for the feature-introducing changes to the CI/CD pipeline, and let this roll as a patch. Thoughts?

Ste1io: > Re the version, I still needed to solve this part of the puzzle. I assume the v7 was also for this, so there was a clean break between appveyor and github

In the past, I've used the rev number to substitute the build number. The workflow run count env variable would work also but only on push to main or version numbers get skipped. Both would need a custom rolled implementation and would need to handle concurrency and rollbacks (or better, a dry run support to prevent eager version bumps).

And both solutions would have to start on a major release.

Using tags would be my suggestion beginning with the v7 release.

asherber: I'm a big fan of nerdbank.gitversioning. I can write more about it later, but it gives you a new build number for each commit, helps with semantic versioning, and doesn't require a new major version to start using it. I use it for all my projects.

Ste1io: +1 @asherber  Never used it personal but I have read up on it and it's been on my list to try out.

asherber: Drop this in the root dir as `Directory.Build.props`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup>
    <PackageReference Include="Nerdbank.GitVersioning" Condition="!Exists('packages.config')">
      <PrivateAssets>all</PrivateAssets>
      <Version>3.6.133</Version>
    </PackageReference>
  </ItemGroup>
</Project>
```

And drop this in the root dir as `version.json`:
```json
{
  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/main/src/NerdBank.GitVersioning/version.schema.json",
  "version": "6.1",
  "publicReleaseRefSpec": [
    "^refs/heads/master$",
    "^refs/heads/v\\d+(?:\\.\\d+)?$"
  ],
  "cloudBuild": {
    "buildNumber": {
      "enabled": true
    }
  }
}
```

nbgv works by calculating the "height" since version.json was changed and using that as the patch value of the version number. So the first version built after you commit these files will be 6.1.1 (height of 1). If you're building from a branch other than master, nbgv assumes it's prerelease and will append the git commit hash. You can also mark a build as explicitly beta by changing the version value in the json file to `6.1-beta`. You can read more about it [here](https://github.com/dotnet/Nerdbank.GitVersioning).

The idea is that if you're just releasing a bugfix, or an internal change like XML docs, just build as per usual and will increment the patch value. If you're adding new functionality, then increment the minor value in version.json when you commit.

pleb:
@Ste1io is this PR ready to merge?

PR #687, opened 2023-08-09
Fix NullReferenceException thrown when executing operations with `dynamic` POCOs

### LINKED ISSUE #667: update dynamic object will cause exception

Ste1io:
The exception occurs when passing in the PocoColumn to the `AddParam` method. As I mentioned in the linked issue thread, this is fine for statically typed POCO objects, but for dynamic or anonymous types, there is no underlying compile-time PropertyInfo for that column (property).

Changing the null check from `if (pc != null)` to `if (pc != null && pc.PropertyInfo != null)` fixes the problem, but I'm not convinced this is the ideal resolution, since the method is specifically designed to accept a null value for the third parameter (for dynamic and anonymous types), and likely there are other bugs related to dynamic that would share the same root cause as this one, somewhere farther down the call path.

https://github.com/CollaboratingPlatypus/PetaPoco/blob/34ff2900ec72d5cae863d4b04f7d87bcc4d15394/PetaPoco/Database.cs#L594-L600

Looking at how anonymous types (which are unaffected, but most likely follow the same call path) are treated internally:

https://github.com/CollaboratingPlatypus/PetaPoco/blob/34ff2900ec72d5cae863d4b04f7d87bcc4d15394/PetaPoco/Database.cs#L756-L757

So the question then is, do we need to be treating dynamic types differently (like anonymous types), or just handle the edge cases with more robust null checks to influence the execution path at runtime? Personally, I've only ever used statically typed objects with PetaPoco, so my knowledge of how runtime types are handled internally is rather limited. Any feedback from @pleb or @asherber would be helpful.

Fixes #667

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,8 +1,13 @@
-Copyright (C) 2011-2023  
 Copyright 2023 CollaboratingPlatypus/PetaPoco
 
-Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the “Software”), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
 
-The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+    http://www.apache.org/licenses/LICENSE-2.0
 
-THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
--- a/PetaPoco.Tests.Integration/Databases/BaseUpdateTests.cs
+++ b/PetaPoco.Tests.Integration/Databases/BaseUpdateTests.cs
@@ -291,23 +291,60 @@ namespace PetaPoco.Tests.Integration.Databases
         }
 
         [Fact]
-        [Trait("Issue", "667")]
-        public void Update_GivenDynamicPoco_ShouldNotThrow() {
-            DB.Insert(_note);
-			var entity = DB.First<dynamic>("SELECT * FROM Note");
-			Should.NotThrow(() => DB.Update("Note", "Id", entity));
+        public void Update_GivenTablePrimaryKeyNameAndDynamicType_ShouldBeValid()
+        {
+            DB.Insert(_person);
+
+            dynamic person = new System.Dynamic.ExpandoObject();
+            person.Id = _person.Id;
+            person.FullName = "Feta";
+            person.Age = 19;
+            person.Dob = new DateTime(1946, 1, 12, 5, 9, 4, DateTimeKind.Utc);
+            person.Height = 190;
+
+            ((int)DB.Update("People", "Id", person)).ShouldBe(1);
+
+            var personOther = DB.Single<Person>(_person.Id);
+            personOther.ShouldNotBe(_person, true);
         }
 
         [Fact]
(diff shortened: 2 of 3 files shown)


Ste1io:
If you feel null checking is sufficient in this case @pleb, commit [50d56e6](https://github.com/CollaboratingPlatypus/PetaPoco/pull/687/commits/50d56e66e4940483be8f88c6012156a4f0d4ac9d) fixes the problem so this PR can be closed.

asherber:
I can't help feeling that update with a dynamic object is a situation that this library is not meant to handle--iow, I would vote for not fixing this. But I'm not going to have time to expand on that for a couple of days.

Ste1io:
That was partly my concern, interested in hearing your more detailed reason when you are able, though, @asherber.

As an alternative fix for this, for now at least, add a check and throw exception like PocoData does for expandoobjects?

Ste1io:
Following up on this, the documentation tests have inline comments specifically stating PetaPoco works well with dynamic objects. Both tests fail with this exception.

What route are we going with here?

Ste1io:
> Looks like the functionality broke in [this commit](https://github.com/CollaboratingPlatypus/PetaPoco/commit/32ea9ca01dc3c5ae2a1a988f43e4aad80619f3fe).

Glad you saw that, was going to point you to the git blame for the previous change but was mobile at the time. Checking `if pi != null`, would accomplish effectively the same as what my PR is doing, since dynamic objects have no compile-time PropertyInfo. Not sure how that ended up inside the DateTime2 PR, but I regardless I do concur with @asherber .

Checking `pc != null` *could* work effectively the same, but only with the additional check I added, or by making the `PropertyInfo` getter inside `ProcoColumn` nullable and handling it accordingly, which isn't possible without a major version bump and targeting newer version of .NET.

asherber:
I added a note to the commit to ask the author what he was trying to accomplish with that particular commit. Let's see if he responds.

Ste1io:
Just saw, cheers.

asherber:
Okay, this was starting to feel familiar to me. And sure enough, look at the last resolved conversation on the [pull request](https://github.com/CollaboratingPlatypus/PetaPoco/pull/630). The author wanted to add `PocoColumn` around in [this commit](https://github.com/CollaboratingPlatypus/PetaPoco/commit/29afd6), and at that point it was my suggestion that he could get rid of `PropertyInfo`. The tests all passed, so it must work, right?

(Side note: These dynamic tests [and possibly others] from the documentation should really be a part of the main test suite, for just this reason.)

Let me go back through things to see if this can be fixed without just reverting that commit; in most cases, it does seem a *little* unfortunate to have to pass around both `pi` and `pc`.

Ste1io:
The tests I added to this PR confirm my fix works.

As for the tests, while I do agree the documentation tests should be part of the main test suite as you mention, they are run by default with all tests are run, so *those should have been caught at the time it was changed*. Still, testing with dynamics, as I did here, really ought to have been a complete test not just an illustration.

asherber:
@Ste1io What's missing from your PR is the magic in line 2187 in the [broken commit](https://github.com/CollaboratingPlatypus/PetaPoco/commit/32ea9ca01dc3c5ae2a1a988f43e4aad80619f3fe). When I suggested to Marcel that `pi` be removed as an argument, I wasn't aware of that bit of code either. I'm looking right now at how that can be brought back without needing to pass both `pi` and `pc`.

Ste1io:
Not at pc so limited by my phones here, but AddParam's logic for dealing with dynamic objects is based on that pi `out` value you mention being null. So it's either send both params and let the null influence the code flow, or add a null check where I did which accomplished the same thing.

Otherwise dynamic objects would need a flag to check which seems unnecessary

pleb:
Loving the work @Ste1io and @asherber. 

My plan was to have the tests run for each PR when I was working on converting from appveyor to GitHub actions. If I get time, I might try working on the Github actions part again

Ste1io:
> Loving the work @Ste1io and @asherber.
> 
> My plan was to have the tests run for each PR when I was working on converting from appveyor to GitHub actions. If I get time, I might try working on the Github actions part again

I've for a motherload of fixes and updates and increased test coverage spread out over roughly a dozen stashes rn @pl

Ste1io (commented):
Perhaps I'm overlooking something obvious, but I'm puzzled by the need to create a pococol if tryget returns false. Dynamic objects can't be used for creating (custom exception thrown by pocodata if you try), and anymous objects are handled specifically to create and cache a pococolumn like a strong type would be. When debugging this, pc has never been null (if it was null in our case here, this bug wouldn't exist right now). It was the combination of pc holding a valid pococolumn with a null propinfo property that upsets things.

In short, I don't believe any of the possible execution paths will result in that branch being taken. If it's an additional safety measure to make it bulletproof for future changes, that makes sense. Not second guessing it, just trying to reason about it.

asherber:
That line of code is only called during updates. If you call one of the overloads like `Update(tableName, primaryKeyName, poco, primaryKeyValue)`, that line makes it possible to add a SQL parameter for the pk even though it's not part of the POCO. I would think the most likely scenario in this case is passing in an anonymous or dynamic POCO. 

Why pass the pk value separately, instead of just make it part of the POCO? Could be different reasons, depending on where the values are coming from in code.

```csharp
Update("MyTable", "ID", new { Name = "Bob" }, 4);

// same as
Update("MyTable", "ID", new { ID = 4, Name = "Bob" });
```

In any case, I'm embarrassed that I didn't catch this change as part of the initial pull request.

Ste1io:
Thanks for the insight, makes more sense. And don't lose any sleep over it; we're all human, and that's the whole point of open source anyways. I'll add an additional test for that tomorrow. Appreciate you taking the time to revisit this.

Ste1io:
Added additional tests using dynamic to match existing anonymous types tests, along with async counterparts.

Branch is ready to merge... Before my phone or my cat do anything else destructive to it lol.

PR #690, opened 2023-09-08
Fix issues with dynamic introduced in 32ea9ca

asherber:
I'd like to suggest this as an alternative to #687. It should fix the issues originally introduced in 32ea9ca by:

1. Restoring in the null check suggested by @Ste1io.
2. Restoring [this code](https://github.com/CollaboratingPlatypus/PetaPoco/blob/29afd6f591934b5a5e9e2c69bb2d9f923cd42577/PetaPoco/Database.cs#L2187) for getting property info for the primary key during an update when there is no `PocoColumn` available in `PocoData`. The deletion of this code was a hidden bug that may also have come into play with #667

### DIFF
--- a/PetaPoco/Database.cs
+++ b/PetaPoco/Database.cs
@@ -599,7 +599,7 @@ namespace PetaPoco
         private void AddParam(IDbCommand cmd, object value, PocoColumn pc)
         {
             // Convert value to from poco type to db type
-            if (pc != null)
+            if (pc?.PropertyInfo != null)
             {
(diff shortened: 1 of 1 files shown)


Ste1io:
That PR is reopened btw, GitHub mobile UI bug in play too blame for the accidental close.

Ste1io:
Could this be merged with the other PR somehow, or you push a commit to that one, so the additional tests are included? I'll probably be at pc for a bit in an hour or so and can test this in docker if you need.

asherber:
Sure, I'll open a PR against your branch -- when you merge that, #687 should update itself.

Issue #691, opened 2023-09-16
PetaPoco. Providers: When using Oracle, calling FetchAsync will result in an exception


jackzhang520:
PetaPoco. Providers: When using Oracle, calling FetchAsync will result in an exception
SQL will start with "SELECT * FROM (SELECT ROW-NUMBER() OVER (ORDER BY (SELECT NULL)) peta_ Rn", but in ORACLE, "SELECT NULL" will prompt ORA-00923: The required FROM keyword was not found. After reading PetaPoco's source code, the BuildPageQuery of OracleDatabaseProvider will be redirected to the BuildPageQuery of SqlServerDatabaseProvider, causing this issue.

Change "SELECT NULL" to " SELECT NULL from DUAL" can fix this issue

Ste1io:
@jackzhang520 is there a specific test that fails from this? I'm currently updating the docker configuration for the test projects, and can look into this once that's done.

jackzhang520:
To fix this,the PetaPoco.Providers.OracleDatabaseProvider need modify:
`public override string BuildPageQuery(long skip, long take, SQLParts parts, ref object[] args)
{
    if (parts.SqlSelectRemoved.StartsWith("*"))
        throw new Exception("Query must alias '*' when performing a paged query.\neg. select t.* from table t order by t.id");

    var helper = (PagingHelper)PagingUtility;
    // when the query does not contain an "order by", it is very slow
    if (helper.SimpleRegexOrderBy.IsMatch(parts.SqlSelectRemoved))
    {
        var m = helper.SimpleRegexOrderBy.Match(parts.SqlSelectRemoved);
        if (m.Success)
        {
            var g = m.Groups[0];
            parts.SqlSelectRemoved = parts.SqlSelectRemoved.Substring(0, g.Index);
        }
    }

    if (helper.RegexDistinct.IsMatch(parts.SqlSelectRemoved))
        parts.SqlSelectRemoved = "peta_inner.* FROM (SELECT " + parts.SqlSelectRemoved + ") peta_inner";

    var sqlPage =
        $"SELECT * FROM (SELECT ROW_NUMBER() OVER ({parts.SqlOrderBy ?? "ORDER BY **(SELECT NULL FROM DUAL)**"}) peta_rn, {parts.SqlSelectRemoved}) peta_paged WHERE peta_rn > @{args.Length} AND peta_rn <= @{args.Length + 1}";
[...]

PR #696, opened 2023-09-20
Containerized tests fixes

### LINKED ISSUE #676: Relocate database agnostic tests to base class
Following up on  https://github.com/CollaboratingPlatypus/PetaPoco/pull/670#issuecomment-1556336751, move tests that aren't db engine specific to appropriate base test class per the test guidelines in the docs. 

Feel free to assign this to me if you like @pleb or @asherber.

Ste1io:
This PR is aimed to get the testing environment to a working state with containerized Docker images. It also includes some spring cleaning in the placement of some tests; however, I've refrained from any potential breaking changes or feature changes (such as fixing support for newer dbms versions not already specifically referenced or supported). I'm also aware that v7 branch already has a significant amount of improvements and changes to the test suites, so I've opted for a few TODO comments vs full-on fixes for most changes (virtual base class methods, updated dbms image versions, etc).

A few remaining tasks I'd like to fix in this PR still; if not, I'll convert them to issues and look into them later; see the commit messages in this PR for more info regarding the fixes.

- [x] setup docker env and update docker-compose.yml
  - requires [Docker Desktop](https://www.docker.com/products/docker-desktop/)
- [x] fix Firebird tests failing with connection error
- [x] fix MySQL stored proc tests failing with error: "Character set 'utf8mb3' is not supported by .Net Framework"
- [x] fix MariaDB tests failing with errors "Object cannot be cast from DBNull to other types",  "Character set 'utf8mb3' is not supported by .Net Framework", and "The given key was not present in the dictionary"
- [x] setup env and verify MsAccess tests
  - requires [MS Access Database Engine 2010 Redistributable](https://www.microsoft.com/en-US/download/details.aspx?id=13255)
- [x] remove MssqlCe tests - past EOL and dropping support in v7
- [x] fix MSAccess paging tests that shouldn't be run for that provider (provider doesn't provide paging, so tests aren't applicable)
- [x] fix MSAccess tests failing from syntax errors
- [x] fix SQLite tests experiencing db locking issues for streaming query tests causing timeouts and test failures.
- [x] documentation-style tests concurrency issues fixed when run in parallel with main test suite
- [x] split shared container images into separate images to isolate the test data from other tests running in parallel (currently causes roughly half of the tests to fail)
- [x] make all base class test functions virtual
- [x] fix `async void` tests to return `Task` to ensure xunit and assertions can propertly catch exceptions and return task completion status
- [x] fix shouldly assertions in async method not using async overload (such as `Should.ThrowAsync`)
- [x] update wiki pages detailing development/contributing, which still contains outdated instructions based on defunct tooling and VS2015 [mostly finished].
- [x] finish moving bugfix repro tests out of derived dbms-specific test classes into base test class

Final tasks related to tests, but to be handled in separate PRs:

- [ ] Create new issue for: add integration test boilerplate classes for oracle; address #691 on that PR
- [ ] Create new issue for: refactor test class inheritance to eliminate duplicate test code between db systems with multiple drivers, and differentiate driver-specific test cases from dbms-specific cases

docker-compose.yml has been updated to version 2. And to version 3. I think that's the latest???...lol

Docker image switched to Firebird 2.5 SuperClassic instead of "latest" to sidestep a plugin mismatch issue with the latest version (v4.0). 2.5 has been EOL for ages, so we should look at 3.0 and 4.0 and add an additional image as a followup.

Fixed a failing test for Firebird DB; the "NULL" keyword is not accepted when adding a new column in an `ALTER TABLE` statement, causing a syntax error. Noticed that many of the `BaseQueryTests` class methods are not marked as virtual, including the offending one here; for consistency, ~~I'd like to see them all virtual, but opted to make that a todo for the moment.~~ I've added this to my list 🤦🏻‍♂️ since half of the MsAccessQueryTests fail due to not having paging support, and need overridden for that provider...may as well just do them all at once and be done with it ig.

Fixed MySQL stored procedure tests that were all failing with "System.NotSupportedException : Character set 'utf8mb3' is not supported by .Net Framework." After a solid hour of searching online to no avail (even ai was no help), I finally tried a "SET NAMES" statement at the top of the sql build script after recalling seeing it in generated db dumps. That fixed it.

Fixed MariaDB tests that were failing with an odd (and rather random) sprinkling of the following two errors:

> System.InvalidCastException : Object cannot be cast from DBNull to other types.
> 
> Stack Trace: 
> IConvertible.ToInt32(IFormatProvider provider)
> Driver.LoadCharacterSets(MySqlConnection connection)
> Driver.Configure(MySqlConnection connection)

> System.NotSupportedException : Character set 'utf8mb3' is not supported by .Net Framework.
> Stack Trace: 
> 
> CharSetMap.GetCharacterSet(DBVersion version, String charSetName)
> MySqlField.SetFieldEncoding()
> NativeDriver.GetColumnData(MySqlField field)

This required specifying mariadb:10.9 for the image, as mentioned in [this SO thread](https://stackoverflow.com/questions/74060289/mysqlconnection-open-system-invalidcastexception-object-cannot-be-cast-from-d), as 10.10 changed the id column type to allow NULL. Doesn't exactly sit well with me, but I supposed this is somewhat the end user's responsibility in a sense. A more robust solution, IMO, would be to switch to the MySqlConnector as the underlying db provider for MariaDB, but that change would need to happen in it's own dedicated PR (and probably qualifies for at least a minor version bump).

Note that after downgrading the image version for MariaDB, including the same "SET NAMES" statement at the top of the build script as MySQL is still necessary to fix the "utf8mb3" charset error (both of them caused by the default encoding changes made by these dbms).

Fixed data integrity bug caused by using the same db image/port for all 4 mssql/mssqlmsdata connections; simultaneous connections running in parallel were causing data to be tainted, and creating countless errors when one test would attempt to cleanup while another was executing.

Documentation tests are now configured to run synchronously; it wasn't worth creating another image just for them, and given how small that group of tests is, whatever impact on the test execution timings is nominal if existent at all.

Fixed sqlite locking issues by running the sqlite in [multithread mode vs the default (serialized)](https://sqlite.org/threadsafe.html). This isn't necessarily the safest design choice obviously, but in the sandboxed environment of the integration tests, concurrency is already being handled by the configured test runner, so there's no need for the additional database-level locking and serialization essentially becomes just a dead weight.

Also included some misc code cleanup, including typo corrections and script consistency improvements as I stumbled across them.

closes #676


Ste1io:
1695224503-PetaPoco_-_Microsoft_Visual_Studio

The two failing doc tests are confirmed to pass with the changes introduced in #687 (not present in this branch).

Ste1io:
Been having some difficulty wrangling the sqlite tests... When run as a batch with other tests, they behave nearly all the time, since there's a slight delay between each test incurred from running all the other parallel tests. If run without other test classes (say, batch testing the query tests), it's pretty much guaranteed to hit a deadlock after half a dozen or so. Haven't had any luck running it as an in-memory file either. As they are now, passing all tests is at the sole mercy of a race condition. 

Feels hacky, but I may need to just have to add manual locking to that class to give the db some breathing room between tests. As it stands, sqlite is just too fragile to be relied on (even for tests that aren't returning a query stream), and once a connection is locked, you're pretty much screwed without restarting the running process. Unacceptable especially if expected to run in an automated workflow.

Any suggestions or ideas?

Curlack:
Can we utilize something similar to what the guys discussed for PHP and
Drupal?

https://www.drupal.org/project/drupal/issues/1120020

They mention starting the connection in IMMEDIATE mode and avoiding
startTransaction exceptions in a try catch.

Ste1io:
> starting the connection in IMMEDIATE mode and avoiding
startTransaction exceptions in a try catch.

Good point, worth a shot in my book. I saw a similar mention to that in a thread on the efcore repo earlier also. I'll try that tomorrow and see what kind of improvements come out of it.

Appreciate the suggestion.

asherber:
Or just configure the SQLite tests not to run in parallel?

Ste1io:
> Or just configure the SQLite tests not to run in parallel?

They aren't running parallel (default behavior of XUnity for all tests from the same class). And running them separate from other tests actually increases the chance of them passing since there's more concurrency involved in the process as a whole.

All integration tests as a single run, usually works without incident, but not every time:
1695339622-PetaPoco_-_Microsoft_Visual_Studio

Testing each Sqlite class in batches, guaranteed lock hit after the first few tests:
1695339733-ShareX_-_Image_viewer

pleb:
Sqlite tests have been dodgy for a while now. I've updated them a few times to work with different sqlite providers. I might get time to have a little look today

Ste1io:
Changing it to run in multi-thread mode instead of serialized mode (default) may actually do the trick...see:

https://sqlite.org/threadsafe.html and https://sqlite.org/compile.html#threadsafe

Ste1io:
Fixed it. Will commit momentarily.
1695340423-PetaPoco_-_Microsoft_Visual_Studio

asherber:
(I was thinking about _across_ test classes, which xunit normally runs in parallel. I didn't realize until I went back to look at the code that the various test classes for each db are defined in a collection, which makes things serial across those classes.)

Curlack:
Nicely done!

Ste1io:
> (I was thinking about _across_ test classes, which xunit normally runs in parallel. I didn't realize until I went back to look at the code that the various test classes for each db are defined in a collection, which makes things serial across those classes.)

Got it, your previous comment does make sense in light of that. Given what you said about how xunit handles collections, would my additional [collection definition class](https://github.com/CollaboratingPlatypus/PetaPoco/blob/54d05327fbf6f814a70d1f9977fb8d2ec1a5c30e/PetaPoco.Tests.Integration/CollectionDefinitions.cs) I made for the doc tests even be necessary? Changing the collection for those four files from "mssql" to "documentation", as I did in https://github.com/CollaboratingPlatypus/PetaPoco/pull/696/commits/54d05327fbf6f814a70d1f9977fb8d2ec1a5c30e, might be sufficient without the xunit attribute?

asherber:
Xunit by default does not run tests within a collection in parallel. By default, again, each class is its own collection. Adding the `[Collection]` attribute to several classes makes them all part of the same collection, so none of the tests in those classes will be run in parallel with each other.

However, Xunit _will_ run the tests of one collection in parallel with the tests in _another_ collection; the collection definition class you added basically tells Xunit not to run that collection in parallel with any other. 

If the issue was that the documentation tests were running in parallel with the MS tests, and failing because the tests were colliding, then your commit should fix it. You might also have chosen just to change the `[Collection]` attribute on the documentation tests to `Mssql` instead of `MssqlTests`. This would have put the tests into the same collection as the other MS tests, and none of those tests would then run in parallel with each other.

Ste1io:
> You might also have chosen just to change the `[Collection]` attribute on the documentation tests to `Mssql` instead of `MssqlTests`. This would have put the tests into the same collection as the other MS tests, and none of those tests would then run in parallel with each other.

As a correction to my last comment, they were previously in "MssqlTests", which differs from the other mssql tests which reside in a "Mssql" collection. Regardless, your point that including in the same collection would also fix the issue confirms what I was asking about.

In this case, keeping them in a separate "Documentation" collection is probably best in this case, since those tests are documenting general library usage, and not specific to mssql aside from that happening to be the provider used when they were written. Ultimately, those tests need revisited anyways (hard-coding sql escapes isn't compatible with all dbms supported by the library, and example test cases really should employ best practices).

Ste1io:
Was there a specific reason for using a MEMO data type for the MS Access db `People`.`Id` column, instead of GUID? The mapped POCO property is `System.GUID`.

https://github.com/CollaboratingPlatypus/PetaPoco/blob/35b5286f66988b5a8c895d2f1eaf85bc40c2b987/PetaPoco.Tests.Integration/Scripts/MSAccessBuildDatabase.sql#L10-L16

TEXT might make sense I suppose, but MEMO seems a bit overkill. Also, MS Access doesn't allow joins on MEMO columns - one of two contributing factors for the two failing tests with this database.

Also, turns out that in addition to changing the data type to GUID, explicitly using the INNER JOIN syntax was necessary for the tests to pass. Even though an implicit JOIN is supposedly no different??? Microsoft is so freaking weird man I swear... Everything they do they gotta do it different just because.

Curlack:
Lol. I know the feeling man. Hang in there.

Ste1io:
> Lol. I know the feeling man. Hang in there.

I really should be used to it by now, after 10 years of putting up with their antics. 🤣

Only one dbms left to go, though, so nearly there. Can't wait to find out what kind of surprises they managed to pack into Compact Edition.

Ste1io:
@pleb I've installed [SSCERuntime_x64-ENU.exe](https://www.microsoft.com/en-US/download/details.aspx?id=30709), and hitting a hard stop from the MssqlCeDBTestProvider constructor: "System.UnauthorizedAccessException : Access to the path 'C:\Windows\assembly\GAC_MSIL\System.Data.SqlServerCe\4.0.0.0__89845dcd8080cc91\sqlceca40.dll' is denied."

https://github.com/CollaboratingPlatypus/PetaPoco/blob/7e6c9401adb011fd8f4b5f07e8210001da529f2b/PetaPoco.Tests.Integration/Databases/MSSQLCe/MssqlCeDBTestProvider.cs#L15-L31

MSSQL Compact Edition is the only dbms I'm unable to test at this point; all remaining tests pass with the changes in this branch via docker and a local MS Access env.

1695527244-PetaPoco_-_Microsoft_Visual_Studio

PetaPoco.Tests.Integration/Databases/BaseQueryLinqTests.cs
```diff
@@ -305,10 +308,8 @@ public virtual async void ExistsAsync_GivenPrimaryKeyMatchingNoRecord_ShouldBeFa
...
         }
 
-        /// <summary>
-        ///     Support the older syntax of starting with a WHERE clause.
-        /// </summary>
-        [Fact]
+        [Fact(DisplayName = "Support the older syntax of starting with a WHERE clause.")]
+        [Trait("Category", "Regression")]
         public virtual async void ExistsAsync_Regression_GivenSqlStringMatchingOneRecord_ShouldBeTrue()
         {
```
Ste1io, line 314: This doesn't look right...shouldn't the regression tests *include* "WHERE" keyword, and the other tests exclude it?

Ste1io: Found the related issue, makes sense now.

Ste1io:
Update on sqlce: I modified the test provider ctor to copy the files into the executable's directory, and it times out during connection attempt.

I've literally **never** messed with SQL Crappy Edition before, and not finding anything very useful online either; getting to the point where idgaf about that provider unless someone jumps in with something super helpful here.

pleb:
Regarding MS access and MS CE, I wouldn't worry too much about making these work in CI/CD. It would be nice, but I suspect few people use PetaPoco against Access and CE. 

Regarding CE, maybe we should remove support as MS themselves no longer support it https://learn.microsoft.com/en-us/lifecycle/products/microsoft-sql-server-compact-40

Ste1io:
Freakin awesome.

Tbf, I'm actually keen on maintaining support for MS Access... With a little bit of fiddling, I should be able to get it to run inside a custom docker setup. And I know people who still use it. Haven't checked on the official status though.

Regardless, I 100% agree that the CI/CD pipeline should be able to run all tests, so ultimately that will be the deciding factor. IMHO, having the tests easily ran on a local developer's workstation with minimal setup via docker is essential also; having to push to remote just to run the tests gets messy real fast. Messier than even my notoriously messy commits 😂.

I'm about half finished with moving tests out to base that aren't dbms-specific. I also added a `BaseTriageTests` class (lmk if you prefer another name) to serve as an initial "catch all" location for bug investigations to go initially when doing a quick repro, prior to becoming regression tests and deciding where they should live long term. Should keep the permanent tests clean and organized, and remove the analysis paralysis from happening when a new contributor wants to open a PR, but is unsure of where it should go or where to start.

Have you had a chance to review everything yet, or any other feedback? I should have my checklist completed by the end of the day; eager to get this in the dev branch so it's not slowing things down anymore.

Ste1io:
The lack of distinction between database systems (such as MSSQL) and drivers/libraries (such as MSSQL MsData) in the tests is resulting in a significant mixture of duplicate code. They already use shared data, any reason not to share tests also? Currently, models are duplicated, all identical tests are duplicated, yet both share the same build scripts and underlying database systems.

It would make more sense to me if drivers inherited from a common database test class (no different from the current setup, except one additional layer of abstraction). This would allow driver-specific tests to be easily differentiated from dbms specific tests, also.

This will make adding tests for Oracle easier as well, especially since there will be a third Oracle driver for v7.

I already implemented this change for MSSQL locally with no issues or test failures. Any reservations on me implementing this change to the integration test suite @pleb @asherber?

asherber:
I agree in general that different drivers for the same dbms should share tests, rather than duplicate.

PetaPoco.Tests.Integration/Databases/Firebird/FirebirdQueryTests.cs
```diff
@@ -39,5 +41,169 @@ public override void Query_ForMultiPocoWithWildcard_ShouldReturnValidPocoCollect
...
+
+            order.Person.ShouldNotBeNull();
+            order.Person.Id.ShouldNotBe(Guid.Empty);
+            order.Person.Name.ShouldStartWith("Peta");
+            order.Person.Age.ShouldBe(18);
+        }
+
+        // FIXME: Firebird does not support multiple statements in a single command? Any solution without resorting to stored procedures or execute blocks?
+        [Fact]
+        public override void QueryMultiple_ForMultiResultsSetWithSinglePoco_ShouldReturnValidPocoCollection()
```
Ste1io, line 115: Is this test case expected to work with Firebird?

PetaPoco.Tests.Integration/Databases/Firebird/FirebirdQueryTests.cs
```diff
@@ -39,5 +41,169 @@ public override void Query_ForMultiPocoWithWildcard_ShouldReturnValidPocoCollect
...
+
+            order.Person.ShouldNotBeNull();
+            order.Person.Id.ShouldNotBe(Guid.Empty);
+            order.Person.Name.ShouldStartWith("Peta");
+            order.Person.Age.ShouldBe(18);
+        }
+
+        // FIXME: Firebird does not support multiple statements in a single command? Any solution without resorting to stored procedures or execute blocks?
+        [Fact]
+        public override void QueryMultiple_ForMultiResultsSetWithMultiPoco_ShouldReturnValidPocoCollection()
```
Ste1io, line 150: Is this test case expected to work with Firebird?

PetaPoco.Tests.Integration/Databases/MSAccess/MsAccessQueryTests.cs
```diff
@@ -171,6 +171,199 @@ public override void Query_ForMultiPocoWithPropertyMissingSetMethod_ShouldThrow(
...
+
+            var person = result.First();
+            person.Id.ShouldNotBe(Guid.Empty);
+            person.Name.ShouldStartWith("Peta");
+            person.Age.ShouldBe(18);
+        }
+
+        // FIXME: Shouldly.ShouldAssertException : result.Count should be 1 but was 0
+        [Fact]
+        public override void QueryMultiple_ForSingleResultsSetWithMultiPoco_ShouldReturnValidPocoCollection()
```
Ste1io, line 202: Not sure what the issue is here...generated SQL is:
```sql
SELECT TOP 1 *
FROM [JoinableOrders] o
INNER JOIN [JoinablePeople] p ON p.[Id] = o.[JoinablePersonId]
WHERE p.[FullName] = @0
ORDER BY 1 DESC;
```

PetaPoco.Tests.Integration/Databases/MSAccess/MsAccessQueryTests.cs
```diff
@@ -171,6 +171,199 @@ public override void Query_ForMultiPocoWithPropertyMissingSetMethod_ShouldThrow(
...
+
+            order.JoinablePerson.ShouldNotBeNull();
+            order.JoinablePerson.Id.ShouldNotBe(0);
+            order.JoinablePerson.Name.ShouldStartWith("Peta");
+            order.JoinablePerson.Age.ShouldBe(18);
+        }
+
+        // FIXME: System.Data.OleDb.OleDbException : Characters found after end of SQL statement.
+        [Fact]
+        public override void QueryMultiple_ForMultiResultsSetWithSinglePoco_ShouldReturnValidPocoCollection()
```
Ste1io, line 256: Not sure what the issue is here...generated SQL is:
```sql
SELECT * FROM [Orders] o
WHERE o.[Id] = @0;
SELECT * FROM [People] p
WHERE p.[FullName] = @1;
```

PetaPoco.Tests.Integration/Databases/MSAccess/MsAccessQueryTests.cs
```diff
@@ -171,6 +171,199 @@ public override void Query_ForMultiPocoWithPropertyMissingSetMethod_ShouldThrow(
...
+
+            order.Person.ShouldNotBeNull();
+            order.Person.Id.ShouldNotBe(Guid.Empty);
+            order.Person.Name.ShouldStartWith("Peta");
+            order.Person.Age.ShouldBe(18);
+        }
+
+        // FIXME: System.Data.OleDb.OleDbException : Characters found after end of SQL statement.
+        [Fact]
+        public override void QueryMultiple_ForMultiResultsSetWithMultiPoco_ShouldReturnValidPocoCollection()
```
Ste1io, line 299: Not sure what the issue is here...generated SQL is:
```sql
SELECT * FROM [JoinableOrders] o
INNER JOIN [JoinablePeople] p ON p.[Id] = o.[JoinablePersonId]
ORDER BY o.[Id] ASC;
SELECT * FROM [JoinableOrderLines] ol
ORDER BY ol.[JoinableOrderId] ASC;
```

PetaPoco.Tests.Integration/Databases/MariaDb/MariaDbQueryTests.cs
```diff
@@ -9,5 +15,169 @@ public MariaDbQueryTests()
...
+
+            order.Person.ShouldNotBeNull();
+            order.Person.Id.ShouldNotBe(Guid.Empty);
+            order.Person.Name.ShouldStartWith("Peta");
+            order.Person.Age.ShouldBe(18);
+        }
+
+        // FIXME: firstOrderLine.Quantity.ToString() should be "1" but was "2"
+        [Fact]
+        public override void QueryMultiple_ForMultiResultsSetWithMultiPoco_ShouldReturnValidPocoCollection()
```
Ste1io, line 124: Seems this should be obvious, but I'm missing what's wrong here. Same assertion failure manifesting on all MySql-based dbms and drivers (MySql, MySql Connector, MariaDB)

PetaPoco.Tests.Integration/Databases/MySQL/MySqlQueryTests.cs
```diff
@@ -9,5 +15,169 @@ public MySqlQueryTests()
...
+
+            order.Person.ShouldNotBeNull();
+            order.Person.Id.ShouldNotBe(Guid.Empty);
+            order.Person.Name.ShouldStartWith("Peta");
+            order.Person.Age.ShouldBe(18);
+        }
+
+        // FIXME: firstOrderLine.Quantity.ToString() should be "1" but was "2"
+        [Fact]
+        public override void QueryMultiple_ForMultiResultsSetWithMultiPoco_ShouldReturnValidPocoCollection()
```
Ste1io, line 124: Seems this should be obvious, but I'm missing what's wrong here. Same assertion failure manifesting on all MySql-based dbms and drivers (MySql, MySql Connector, MariaDB)

PetaPoco.Tests.Integration/Databases/MySQLConnector/MySqlConnectorQueryTests.cs
```diff
@@ -9,5 +15,169 @@ public MySqlConnectorQueryTests()
...
+
+            order.Person.ShouldNotBeNull();
+            order.Person.Id.ShouldNotBe(Guid.Empty);
+            order.Person.Name.ShouldStartWith("Peta");
+            order.Person.Age.ShouldBe(18);
+        }
+
+        // FIXME: firstOrderLine.Quantity.ToString() should be "1" but was "2"
+        [Fact]
+        public override void QueryMultiple_ForMultiResultsSetWithMultiPoco_ShouldReturnValidPocoCollection()
```
Ste1io, line 124: Seems this should be obvious, but I'm missing what's wrong here. Same assertion failure manifesting on all MySql-based dbms and drivers (MySql, MySql Connector, MariaDB)

Curlack:
SELECT * FROM [Orders] o
WHERE o.[Id] = @0;
SELECT * FROM [People] p
WHERE p.[FullName] = @1;

I think the error is due to the text after semicolon.

Ste1io:
> I think the error is due to the text after semicolon.

Meaning, same problem as Firebird? eg: only one statement allowed per command?

PetaPoco.Tests.Integration/Databases/MSAccess/MsAccessQueryTests.cs
Ste1io: Following @Curlack's [comment](https://github.com/CollaboratingPlatypus/PetaPoco/pull/696#issuecomment-1737835299), and some [follow-up](https://stackoverflow.com/questions/1280021/in-ms-access-is-it-possible-to-execute-multiple-queries) [research](https://stackoverflow.com/questions/770614/running-multiple-sql-statements-in-the-one-operation), seems MS Access doesn't accept multiple statements in a single executed command in the context of standalone SQL queries (as supposed to VBA procedure or macro).

PetaPoco.Tests.Integration/Databases/MSAccess/MsAccessQueryTests.cs
Ste1io: Following @Curlack's [comment](https://github.com/CollaboratingPlatypus/PetaPoco/pull/696#issuecomment-1737835299), and some [follow-up](https://stackoverflow.com/questions/1280021/in-ms-access-is-it-possible-to-execute-multiple-queries) [research](https://stackoverflow.com/questions/770614/running-multiple-sql-statements-in-the-one-operation), seems MS Access doesn't accept multiple statements in a single executed command in the context of standalone SQL queries (as supposed to VBA procedure or macro).

PetaPoco.Tests.Integration/Databases/MySQLConnector/MySqlConnectorQueryTests.cs
Ste1io: Consolidating this thread to the [MySQL](https://github.com/CollaboratingPlatypus/PetaPoco/pull/696#discussion_r1338988806) comments since they stem from the same issue.

PetaPoco.Tests.Integration/Databases/MySQL/MySqlQueryTests.cs
Ste1io: I'm not sure *how* this could be happening, but this test passes on the MySQL instance in my local dev env (v8.0.25). In the Docker image (latest: v8.1.0), it fails. Manually inspecting a few records from `OrderLines`:

```sql
`SELECT * FROM `OrderLines` ol ORDER BY ol.`OrderId` ASC;`
```

| # Id | OrderId | Qty | Status | SellPrice |
|------|---------|-----|--------|-----------|
| '1'  | '1'     | '1' | '0'    | '9.9900'  |
| '2'  | '1'     | '2' | '0'    | '19.9800' |
| '3'  | '2'     | '1' | '0'    | '9.9900'  |
| '4'  | '2'     | '2' | '0'    | '19.9800' |

It occurred to me since we're only sorting based on `OrderId`, and since there are two records for each unique `OrderId`, it would be possible I guess that a valid ordering could also be:

| # Id | OrderId | Qty | Status | SellPrice |
|------|---------|-----|--------|-----------|
| '2'  | '1'     | '2' | '0'    | '19.9800' |
| '1'  | '1'     | '1' | '0'    | '9.9900'  |
| '4'  | '2'     | '2' | '0'    | '19.9800' |
| '3'  | '2'     | '1' | '0'    | '9.9900'  |

Which, technically, isn't wrong, but most certainly does qualify as undefined behavior.

I'm still mystified as to why it's only exhibited on MySQL DBMS on the Docker image, and not my local (or the other DBMS instances on Docker either, for that behavior).

Tests now pass, with a modified ORDER BY:

```sql
SELECT * FROM `OrderLines` ol ORDER BY ol.`Id` ASC, ol.`OrderId` ASC;
```

Still rather unsettling as I don't know what's causing the sorting to consistently differ between db instances.

https://github.com/CollaboratingPlatypus/PetaPoco/blob/c2dbb11c8fc5c10938f577acbcc57b22761b9fd7/PetaPoco.Tests.Integration/Databases/MySQLConnector/MySqlConnectorQueryTests.cs#L140

PetaPoco.Tests.Integration/Databases/MySQLConnector/MySqlConnectorQueryTests.cs
Ste1io: [Resolved](https://github.com/CollaboratingPlatypus/PetaPoco/pull/696#discussion_r1339138316) with same fix as MySql.

PetaPoco.Tests.Integration/Databases/MariaDb/MariaDbQueryTests.cs
Ste1io: [Resolved](https://github.com/CollaboratingPlatypus/PetaPoco/pull/696#discussion_r1339138316) with same fix as MySql.

Ste1io:
@pleb 
1wlzuo

Curlack:
It seems so. Saw the same problem here:
https://www.codeproject.com/Questions/1137585/How-do-I-select-data-from-multiple-table-in-single
They solved it using DataAdapter.Fill, not sure if we can utilize it
somehow. Otherwise, yes, same as Firebird, although I'm not familiar with
it 😁

You've done one hell of a job with this code base, well done man! I'm
thinking of setting everything up on my side and jump in there with you
whenever I can get some spare time...

Ste1io:
> I'm thinking of setting everything up on my side and jump in there with you whenever I can get some spare time...

You should... Will be much easier once this gets merged. Currently working on a table data delete feature to add to tests that maps out the constraints from the sql meta and deletes the data bottom-up. Not having to run the build scripts with a total teardown and create for each test should speed up the tests significantly.

And been itching to add some features I've been using myself for awhile now, but couldn't really bring to the table without functioning tests. Like my PocoDataHelper for typesafe inline escaping with intellisense support, with next to zero overhead (no expression compilation and built-in caching).

Appreciate the support and input btw. First time messing with MS Access, CE, Firebird, and Postgres. Always fun to figure out new systems you hope you'll never use again 😂 though I got to say, very impressed with Postgres's performance and infra; definitely going to mess around with it more.

I'll revisit the data sets and DataAdapter.Fill mentioned in your link. It was my understanding that only worked with VBA, but I could be wrong.

pleb:
Love your work, @Ste1io

PR #707, opened 2023-10-22
Fix some more failing integration tests

Curlack:
Would've like to separate the core changes into a different PR, but got ahead of myself.

Issue #699, opened 2023-09-30
Add integration testing for Oracle, for official DBMS support


Ste1io:
PetaPoco has "unofficially" supported Oracle for quite some time now... I'd like to implement the integration tests for Oracle so that can change finally, and we can claim official support without a big *"but"* after. This will also allow us to confidently investigate, address, and close multiple issues that are unclosed, and can most likely be readily fixed provided tests are possible.

Happy to handle this myself alongside the remaining test refactoring.

Closes #691
Addresses #626
Addresses #613

Curlack:
Busy setting up everything. Have used Oracle for a while couple of years
back. Might be ready to lend a hand soon 😁

Ste1io:
Let me know how setup goes with your local environment, @Curlack, and if you encounter anything you think should be added to the updated wiki page. 

Given your background in Oracle, I don't mind passing this one over to you if you want, or collaborating together on it, whichever is best for you.

Curlack:
Setup was smooth. After following all the instructions. The code compiled
and all tests passed 😎.
Collaborating at first at least. Need to find my bearings

Curlack:
I'm gonna need the docker setup for oracle. I can see a comment for oracle pointing to "sath89/oracle-12c:latest" image, but to be honest, I have no clue how all the docker stuff works. Will figure it out as I go. Not sure if we'll start with 12c and go up to 23c or for all intents and purposes 12c will always be enough.

PS: Also wanted to ask if this will be our primary medium of communication and (since we're 6 hours apart) what times would suite you best?

Ste1io:
For the docker images, most of it's been trial and error and checking docs... the sathu89 image (that commented service predates my recent PR) is no longer available when I checked (apparently due to a dmca from Oracle ~2019). afaik, 19c is current LTS, so that would be my suggestion. Including 12c for backwards compatibility with the older Oracle provider (the orphan in the root directory, I suppose) may be necessary.

Matrix testing hasn't really been brought up yet, but you'll notice I added comments about firebird also as the current image used for it is a couple major versions behind, like 12c.

Just noticed I neglected to update the .vscode extensions with it, but if you install the the [Docker extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) for VSCode, testing images is fairly straightforward simply by right-clicking on the compose.yml after making your changes and selecting the services you want to fire up.

Curlack:
[...]

PR #709, opened 2023-10-29
Add support for case insensitive dynamic pocos

Curlack:
Split Oracle scripts and tests to test usage of "ordinary" and "delimited" identifiers.

Add support for case insensitive dynamic objects. `ExpandoPoco` vs `ExpandoObject`. To be used by `OracleOrdinaryDatabaseProvider`.

Add `UseOrdinaryIdentifiers` flag to `IProvider` to provide a global means of controlling the activation of case insensitive dynamic objects.

Add `ignoreCase` optional parameter to all Database constructors (except the ones taking in a provider) and Poco factory methods.

Add `UsingCaseInsensitiveProvider` configuration extension.

Switch between `OracleDatabaseProvider` and `OracleOrdinaryDatabaseProvider` based on `UseOrdinaryIdentifiers` flag when Resolving Provider.

### DIFF
--- a/PetaPoco.Tests.Integration/Databases/DatabaseTests.cs
+++ b/PetaPoco.Tests.Integration/Databases/DatabaseTests.cs
@@ -36,12 +36,14 @@ namespace PetaPoco.Tests.Integration
         public virtual void Construct_GivenConnection_ShouldBeValid()
         {
             var factory = DB.Provider.GetFactory();
+            var ignoreCase = DB.Provider.UseOrdinaryIdentifiers;
+
             using (var connection = factory.CreateConnection())
             {
                 connection.ConnectionString = DB.ConnectionString;
                 connection.Open();
 
-                using (var db = new Database(connection))
+                using (var db = new Database(connection, ignoreCase: ignoreCase))
                 {
                     AfterDbCreate(db);
                     var key = db.Insert(_note);
@@ -57,8 +59,9 @@ namespace PetaPoco.Tests.Integration
         {
             var providerName = ProviderName;
             var connectionString = DB.ConnectionString;
+            var ignoreCase = DB.Provider.UseOrdinaryIdentifiers;
 
-            using (var db = new Database(connectionString, providerName))
+            using (var db = new Database(connectionString, providerName, ignoreCase: ignoreCase))
             {
                 AfterDbCreate(db);
                 var key = db.Insert(_note);
@@ -73,8 +76,9 @@ namespace PetaPoco.Tests.Integration
         {
             var factory = DB.Provider.GetFactory();
             var connectionString = DB.ConnectionString;
+            var ignoreCase = DB.Provider.UseOrdinaryIdentifiers;
 
-            using (var db = new Database(connectionString, factory))
+            using (var db = new Database(connectionString, factory, ignoreCase: ignoreCase))
             {
                 AfterDbCreate(db);
                 var key = db.Insert(_note);
@@ -90,8 +94,9 @@ namespace PetaPoco.Tests.Integration
         {
             var connectionString = DB.ConnectionString;
             var entry = ConfigurationManager.ConnectionStrings.Cast<ConnectionStringSettings>().FirstOrDefault(c => c.ConnectionString.Equals(connectionString));
+            var ignoreCase = DB.Provider.UseOrdinaryIdentifiers;
 
-            using (var db = new Database(entry.Name))
+            using (var db = new Database(entry.Name, ignoreCase: ignoreCase))
             {
                 AfterDbCreate(db);
                 var key = db.Insert(_note);
--- a/PetaPoco.Tests.Integration/Databases/OracleTests/OracleDatabaseTests.cs
+++ b/PetaPoco.Tests.Integration/Databases/OracleTests/OracleDatabaseTests.cs
@@ -3,12 +3,29 @@ using Xunit;
 
 namespace PetaPoco.Tests.Integration.Databases.Oracle
 {
-    [Collection("Oracle")]
-    public class OracleDatabaseTests : DatabaseTests
+    public abstract partial class OracleDatabaseTests : DatabaseTests
     {
-        public OracleDatabaseTests()
-            : base(new OracleTestProvider())
+        protected OracleDatabaseTests(TestProvider provider)
+            : base(provider)
         {
+        }
+
+        [Collection("Oracle.Delimited")]
+        public class Delimited : OracleDatabaseTests
+        {
+            public Delimited()
+                : base(new OracleDelimitedTestProvider())
+            {
+            }
+        }
+
+        [Collection("Oracle.Ordinary")]
+        public class Ordinary : OracleDatabaseTests
+        {
+            public Ordinary()
+                : base(new OracleOrdinaryTestProvider())
+            {
+            }
         }
     }
 }
--- a/PetaPoco.Tests.Integration/Databases/OracleTests/OracleDeleteTests.cs
+++ b/PetaPoco.Tests.Integration/Databases/OracleTests/OracleDeleteTests.cs
@@ -3,12 +3,29 @@ using Xunit;
 
 namespace PetaPoco.Tests.Integration.Databases.Oracle
 {
-    [Collection("Oracle")]
-    public class OracleDeleteTests : DeleteTests
+    public abstract partial class OracleDeleteTests : DeleteTests
     {
-        public OracleDeleteTests()
-            : base(new OracleTestProvider())
+        protected OracleDeleteTests(TestProvider provider)
+            : base(provider)
         {
+        }
+
+        [Collection("Oracle.Delimited")]
+        public class Delimited : OracleDeleteTests
+        {
+            public Delimited()
+                : base(new OracleDelimitedTestProvider())
+            {
+            }
+        }
+
+        [Collection("Oracle.Ordinary")]
+        public class Ordinary : OracleDeleteTests
+        {
+            public Ordinary()
+                : base(new OracleOrdinaryTestProvider())
+            {
+            }
         }
     }
 }
--- a/PetaPoco.Tests.Integration/Databases/OracleTests/OracleExecuteTests.cs
+++ b/PetaPoco.Tests.Integration/Databases/OracleTests/OracleExecuteTests.cs
@@ -3,12 +3,29 @@ using Xunit;
 
 namespace PetaPoco.Tests.Integration.Databases.Oracle
 {
-    [Collection("Oracle")]
-    public class OracleExecuteTests : ExecuteTests
+    public abstract partial class OracleExecuteTests : ExecuteTests
     {
-        public OracleExecuteTests()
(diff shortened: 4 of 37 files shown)


Ste1io:
@Curlack, looking forward to taking a look at the changes. There is a slight issue with commits from the previous PR being included in the new one, likely due to the hard reset earlier (easily fixed, though). Looking at the commit hashes in this case, it appears that the new commits were done without first rebasing onto upstream's feature/oracle-support branch following the final force-push you did from the Github web UI (notice commit e57bf10 in this PR, which was actually replaced by b36c22b when you did the final force-push in #707).

Should be fairly painless to fix, you probably can just c/p the git commands below all at once and call it a day. I added inline comments so you know what's going on, but essentially we're just making a new branch without your new changes, applying only your new commits that we want to it, then replacing your original feature branch with the new one.

```powershell
# Create a new private feature branch from upstream's "feature/oracle-support" branch; this gives you a clean slate based on the upstream branch without any of your new commits
git checkout -b fix-oracle-quoting upstream/feature/oracle-support
# Now we apply your changes for this PR (in order) to this new private feature branch; this is effectively the same as what rebasing does, by replaying the commits from another branch as if they all occurred in order on a single branch.
git cherry-pick 82f67b8 5cf2dea df1d8b2 0addb74
# In this case, we need to fix up your "feature/oracle-support" private branch which you opened the PR from, so just overwrite it with our new feature branch "fix-oracle-quoting"
git branch -f feature/oracle-support fix-oracle-quoting
# Force push to your remote and the PR should be fixed
git push origin feature/oracle-support --force
```

For what it's worth, git can be a bit of a brainf**k between local, remote, origin, upstream, and who is mine and theirs LOL, so don't feel bad.

I generally I create a short-lived private feature branch for each PR; it's not necessary, but it does encapsulate that PR's changes separately from the target branch fairly well.

- Once the PR is approved, I can delete that private feature branch, sync my fork with upstream (same as *merging upstream's changes into my branch*), and create a new private feature branch for the next PR.
- While working on my short-lived feature branch, if there are changes upstream to the branch I intend to open my PR against, I just *rebase onto upstream's branch*, that way those commits get applied to my feature branch alongside my own.

Curlack:
Thanks for all the help and patience. I'm busy trying to fix as you indicated, but it seems git doesn't know about branch `upstream/feature/oracle-support`. Is there something else I'm missing?

asherber:
I've been following this from the sidelines, but looking now at the code changes, I have reservations about what's being added -- seems like a lot of things that shouldn't really be the business of this library.

Many databases have case-sensitivity issues, and I think it's up to the user to set up their default mapper to handle it. I don't like adding an `ignoreCase` parameter in the `Database` constructors (and the new extension method) when that parameter only has an effect for Oracle.

asherber:
Also, adding a property to `IProvider` is a breaking change.

Ste1io:
> Thanks for all the help and patience. I'm busy trying to fix as you indicated, but it seems git doesn't know about branch `upstream/feature/oracle-support`. Is there something else I'm missing?

I'm on mobile so double check for typos... Make sure you fetch upstream first. Also this repo needs set as upstream remote if it isn't already; check with `git remote -v`. If none set, `git remote add upstream https://github.com/CollaboratingPlatypus/PetaPoco.git`.

Run `git fetch upstream`, then you should be good.

Curlack:
> I've been following this from the sidelines, but looking now at the code changes, I have reservations about what's being added -- seems like a lot of things that shouldn't really be the business of this library.
> 
> Many databases have case-sensitivity issues, and I think it's up to the user to set up their default mapper to handle it. I don't like adding an `ignoreCase` parameter in the `Database` constructors (and the new extension method) when that parameter only has an effect for Oracle.



> Also, adding a property to `IProvider` is a breaking change.

Noted. I was expecting this, hence me mentioning the eyebrows earlier.
Just wanted to somehow demonstrate one way of solving the dynamic case sensitivity issue and also how it ties into the database, provider and factories. Out of all the things I tried, this seemed the cleanest, believe it or not.

Open to suggestions at this point, but completely agree with you. Of the two extremes, I first went [this route](https://github.com/CollaboratingPlatypus/PetaPoco/issues/699#issuecomment-1778207981) then we'll work our way back to the [ideal](https://github.com/CollaboratingPlatypus/PetaPoco/issues/699#issuecomment-1778314360).

Curlack:
> > Thanks for all the help and patience. I'm busy trying to fix as you indicated, but it seems git doesn't know about branch `upstream/feature/oracle-support`. Is there something else I'm missing?
> 
> I'm on mobile so double check for typos... Make sure you fetch upstream first. Also this repo needs set as upstream remote if it isn't already; check with `git remote -v`. If none set, `git remote add upstream https://github.com/CollaboratingPlatypus/PetaPoco.git`.
> 
> Run `git fetch upstream`, then you should be good.

I think I did it! ...,but you'll be the judge when you get a chance.

PS: When you do the review, don't review with the aim to approve the PR this time, but rather as a way to show you my changes to better collaborate. Not sure if there is another way to share code that does involve a PR.

Ste1io:
> I've been following this from the sidelines, but looking now at the code changes, I have reservations about what's being added -- seems like a lot of things that shouldn't really be the business of this library.
> 
> Many databases have case-sensitivity issues, and I think it's up to the user to set up their default mapper to handle it. I don't like adding an `ignoreCase` parameter in the `Database` constructors (and the new extension method) when that parameter only has an effect for Oracle.

Retaining case sensitivity is a concern I share also. I don't think it has to be compromised here though... More below...

Having a `ShouldQuote` bool at a"global" level disabled for default Oracle configuration, sounds reasonable to me if it doesn't have an adverse affect on usability and performance for existing dbs. But realistically, there's got to be a per-table/column override (which I can see implemented like the ignore column flag, or explicit columns etc). Probably nauseating how many dbs out there have mixed quoted tables and columns... Adding an optional column info property should be possible without creating any regression.

I'm not keen on adding overloads for literally half the API with a `shouldQuote` param to apply to just that one call (commands that accept dynamics, anon, sqlstrings identifying takes or columns) would all be candidates. *shucks*

Ste1io:
Welp, replying on my phone, clearly haven't learned my lesson... Disregard the accidental PR status change. I'll be at pc in a couple hrs, going to refrain from further participation until then for the sanity of all of you lol.

asherber:
@Curlack I may be missing something, but the fact that `dynamic` objects are not inherently case-insensitive isn't an issue for this library. PetaPoco maps database values to POCOs, and I think that end users could map things to a case-insensitive `dynamic` with current functionality, if they wanted. At most, it would take a few lines in an extension method. Providing a case-insensitive dynamic POCO seems outside the scope of this library.

I've never used Oracle, so there are probably some things specific to that db that I'm unaware of. Reading your [comment](https://github.com/CollaboratingPlatypus/PetaPoco/issues/699#issuecomment-1773893583) about delimited vs. ordinary, it sounds similar to how [Postgres](https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) handles things, which I am familiar with. IMO, it's the job of the user to understand that PP always escapes identifiers and plan accordingly with respect to database design and PetaPoco mapper. For example, since Postgres folds unquoted names to lowercase, it's common to name tables and columns all in lowercase. But since I know that PetaPoco will quote all identifiers, and since POCO class and property names are PascalCase, I always attach a mapper to my `Database` object that similarly lowercases table and column names, since they will be quoted and hence case-sensitive. Alternatively, I could have used a custom provider that makes `EscapeSqlIdentifier` a no-op, which I think you discuss elsewhere.

I'll also add that an option on the provider to "ignore case" seems vague. For example, queries against a `text` or `varchar` column in Postgres use case-sensitive matching. If I tell PetaPoco to ignore case, will it make these queries case-insensitive as well? 

(This is also something that can be solved with current functionality. I defined my columns as `citext`, which is a case-insensitive text type in Postgres, but since the incoming parameter also has to be `citext` in order for this to work, I added a handler to `OnCommandExecuting` in my `Database` object that loops through query params and changes the type of all string params to `citext`.)

So...I don't see anything here that adds relevant functionality to the library, and I would currently vote against merging this PR. If you can explain succinctly what I'm not seeing, I'm open to having my mind changed.

Curlack:
> Providing a case-insensitive dynamic POCO seems outside the scope of this library.

Agreed.


> I'll also add that an option on the provider to "ignore case" seems vague

Couldn't think of a better name, but mentioned in the XML comment that it's whether to use case insensitive dynamic object. This only applies to the property accessors in the dynamic object coming from the `PocoData` factory for `typeof(object)` e.g. Oracle still gives `{"ID": 1}`, but we can get/set `poco.Id`.


> I don't see anything here that adds relevant functionality to the library, and I would currently vote against merging this PR.

Your comments have been very helpful and actually gave me a better idea. So agreed, no merge.

#### Summary

From Oracle's perspective, delimited identifiers are the exception to the rule. My approach was based on this, which led me down the wrong path. From PP's perspective, ordinary identifiers are actually the exception to the rule. For my next trick, I'll be taking this approach. I'm gonna have a look at some of the Postgres tests and core implementation and start reworking these changes. @Ste1io don't even bother reviewing this PR.

**The changes are too invasive with no real value to the core concept of this library. They also contain breaking changes with no real justification**

#### New game plan:
- I will start "vanilla" i.e. manually remove violating changes based on git diffs. I will keep the new Oracle integration test split and just move the 'OracleOrdinaryDatabaseProvider' into the integration tests project, as a custom provider (for the "Ordinary" slice of the Oracle tests).
- See how many "Delimted" tests fail and tweak where necessary to make them pass. These tests should at least be supported by the library out of the box (except for the ones using `dynamic`).
- Take another crack at the "Ordinary" slice and see where we're at as far as failing tests are concerned.
- Provide missing functionality to custom provider and custom mapper, to get as many tests to pass as is possible.
- Modify remaining failing tests to demonstrate deviations/workarounds required to arrive at the destination.
- Generate all scenarios around casing for Oracle vs C# and post them in the original issue.

Thanks again for all the help, really appreciate it.

asherber:
@Curlack Thanks for your comments. I haven't looked in detail at the Oracle code, but from a high level, I would suggest focusing on getting current tests to pass out of the box. Maybe there's something that needs to be restructured in the tests in order to accommodate Oracle, though I'm pretty sure there was a point at which they all passed. Maybe take a look at [NPoco](https://github.com/schotime/NPoco) to see how they do things. (NPoco is a well-maintained fork of the project this repo also came from.)

If you think there's a need to provide alternative approaches for Oracle, consider whether the better place for that might be a separate library or even just a page in the Wiki with tips and hints. Consider also whether Oracle is more in need of this attention than other databases; I don't see any open issues in the repo relating to these things.

kchanghui:
PetaPoco 对postgres 数据库支持有bug  ,  当插入数据表,  数据表会带有引号“myTable”,  报错信息： Npgsql.PostgresException:“42P01: 关系 "myTable" 不存在”；  实际上表是存在的，查询没问题， 当执行insert，update ， delete ,就会包这样的错误

PR #710, opened 2023-11-01
Fix remaining failing integration tests

### LINKED ISSUE #517: PocoData.ForType() caching
When I call `PocoData.ForType(type, mapper)` for the first time, the resulting `PocoData` object is stored in the cache keyed only by the type. This means that if I later call `PocoData.ForType(type, otherMapper)`, I get back the initial PD object, which was created with a different mapper.

This is probably something of an edge case, but imagine a situation where a single type is used for queries in two different databases, each with its own mapper. The first database to use the class will cache its PD, and then both databases will always get that object.

I think it would be better if the cache had a composite key of type and mapper type. (Even then, it's possible for an instance of a mapper to have its properties changed between uses, but I would think that's even rarer.)

I'm happy to work on a PR for this.

Curlack:
Test Oracle Delimited and Ordinary Identifiers Separately.

#### IntegrationTests
Split Oracle DB scripts, connection strings and tests in two.
One set of tests to use PP OOTB (Delimited) and one set of tests to use PP with Custom provider and mapper (Ordinary).
Fix Oracle setup script. Wait for tablespace drop operation to complete before creating the new tablespace.

#### Core
**Database.cs**
- Escape Oracle Stored Procedure parameter names instead of ensuring paramater prefix.

**OracleDatabaseProvider.cs**
- Quote identifier when escaping (no Regex or ToUpperInvariant).
- Disable statement caching for Oracle client.

### DIFF
--- a/PetaPoco.Tests.Integration/Databases/OracleTests/OracleUpdateTests.cs
+++ b/PetaPoco.Tests.Integration/Databases/OracleTests/OracleUpdateTests.cs
@@ -1,14 +1,114 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using PetaPoco.Core;
+using PetaPoco.Tests.Integration.Models;
 using PetaPoco.Tests.Integration.Providers;
+using Shouldly;
 using Xunit;
 
 namespace PetaPoco.Tests.Integration.Databases.Oracle
 {
-    [Collection("Oracle")]
-    public class OracleUpdateTests : UpdateTests
+    public abstract partial class OracleUpdateTests : UpdateTests
     {
-        public OracleUpdateTests()
-            : base(new OracleTestProvider())
+        protected OracleUpdateTests(TestProvider provider)
+            : base(provider)
         {
+        }
+
+        [Collection("Oracle.Delimited")]
+        public class Delimited : OracleUpdateTests
+        {
+            public Delimited()
+                : base(new OracleDelimitedTestProvider())
+            {
+            }
+        }
+
+        [Collection("Oracle.Ordinary")]
+        public class Ordinary : OracleUpdateTests
+        {
+            public Ordinary()
+                : base(new OracleOrdinaryTestProvider())
+            {
+            }
+
+            [Fact]
+            [Trait("Issue", "#667")]
+            public override void Update_GivenTablePrimaryKeyNameAndDynamicType_ShouldNotThrow()
+            {
+                var pd = PocoData.ForType(typeof(Note), DB.DefaultMapper);
+                var tblNote = DB.Provider.EscapeTableName(pd.TableInfo.TableName);
+
+                DB.Insert(note);
+                var entity = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();
+
+#if !NETCOREAPP
+                Should.NotThrow(() => DB.Update("NOTE", "ID", (object)entity));
+#else
+                Should.NotThrow(() => DB.Update("NOTE", "ID", entity));
+#endif
+            }
+
+            [Fact]
+            public override void Update_GivenTablePrimaryKeyNameAndDynamicType_ShouldUpdate()
+            {
+                var pd = PocoData.ForType(typeof(Note), DB.DefaultMapper);
+                var tblNote = DB.Provider.EscapeTableName(pd.TableInfo.TableName);
+
+                DB.Insert(note);
+                var entity = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();
+
+                entity.TEXT += " was updated";
+
+#if !NETCOREAPP
+                DB.Update("NOTE", "ID", (object)entity);
+#else
+                DB.Update("NOTE", "ID", entity);
+#endif
+
+                var entity2 = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();
+                ((string)entity2.TEXT).ShouldContain("updated");
+            }
+
+            [Fact]
+            [Trait("Issue", "#667")]
+            public override async Task UpdateAsync_GivenTablePrimaryKeyNameAndDynamicType_ShouldNotThrow()
+            {
+                var pd = PocoData.ForType(typeof(Note), DB.DefaultMapper);
+                var tblNote = DB.Provider.EscapeTableName(pd.TableInfo.TableName);
+
+                await DB.InsertAsync(note);
+                var entity = (await DB.FetchAsync<dynamic>($"SELECT * FROM {tblNote}")).First();
+
+                // https://docs.shouldly.org/documentation/exceptions/throw#shouldthrowfuncoftask
+#if !NETCOREAPP
+                Should.NotThrow(async () => await Task.Run(() => DB.UpdateAsync("NOTE", "ID", (object)entity)));
+#else
+                await Should.NotThrowAsync(() => DB.UpdateAsync("NOTE", "ID", entity));
+#endif
+            }
+
+            [Fact]
+            public override async Task UpdateAsync_GivenTablePrimaryKeyNameAndDynamicType_ShouldUpdate()
+            {
+                var pd = PocoData.ForType(typeof(Note), DB.DefaultMapper);
+                var tblNote = DB.Provider.EscapeTableName(pd.TableInfo.TableName);
+
+                await DB.InsertAsync(note);
+                var entity = (await DB.FetchAsync<dynamic>($"SELECT * FROM {tblNote}")).First();
+
+                entity.TEXT += " was updated";
+
+#if !NETCOREAPP
+                await DB.UpdateAsync("NOTE", "ID", (object)entity);
+#else
+                await DB.UpdateAsync("NOTE", "ID", entity);
+#endif
+
+                var entity2 = (await DB.FetchAsync<dynamic>($"SELECT * FROM {tblNote}")).First();
+                ((string)entity2.TEXT).ShouldContain("updated");
+            }
         }
     }
 }
--- a/PetaPoco.Tests.Integration/Databases/UpdateTests.cs
+++ b/PetaPoco.Tests.Integration/Databases/UpdateTests.cs
@@ -37,7 +37,7 @@ namespace PetaPoco.Tests.Integration
             Name = "Peta"
         };
 
-        private Note _note = new Note
+        protected Note note = new Note
         {
             CreatedOn = new DateTime(1948, 1, 11, 4, 2, 4, DateTimeKind.Utc),
             Text = "Peta's Note",
@@ -80,7 +80,6 @@ namespace PetaPoco.Tests.Integration
             orderLine.SellPrice = 5.99m;
             orderLine.Status = OrderLineStatus.Allocated;
         }
-
         #endregion
 
         [Fact]
@@ -376,7 +375,7 @@ namespace PetaPoco.Tests.Integration
             var pd = PocoData.ForType(typeof(Note), DB.DefaultMapper);
             var tblNote = DB.Provider.EscapeTableName(pd.TableInfo.TableName);
 
-            DB.Insert(_note);
+            DB.Insert(note);
             var entity = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();
 
 #if !NETCOREAPP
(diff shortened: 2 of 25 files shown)


asherber:
I know this PR is only going to an internal branch right now, not to develop, but I think the changes to `PocoData` should be their own PR -- and should stay that way going into develop. That's too significant a plumbing change to be buried in a PR about Oracle tests. Best way to handle is probably to do that on its own in a separate branch, and then once its merged to develop merge develop into this branch.

PetaPoco/Core/PocoData.cs
```diff
@@ -148,8 +149,16 @@ public static PocoData ForObject(object obj, string primaryKeyName, IMapper defa
...
-                    if (col != primaryKeyName)
-                        pd.Columns.Add(col, new ExpandoColumn() { ColumnName = col });
+                    if (col == primaryKeyName) continue;
+                    if (pd.Columns.TryGetValue(col, out PocoColumn pc))
+                    {
+                        //When col and primaryKeyName only differ by case,
+                        //we need ColumnName to match case exactly.
+                        pc.ColumnName = col;
+                        continue;
+                    }
```
asherber, line 159: What's going on here?

Curlack: Consider the following test code:
```c#
var pd = PocoData.ForType(typeof(Note), DB.DefaultMapper);
var tblNote = DB.Provider.EscapeTableName(pd.TableInfo.TableName);

CreateNote();
var entity = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();

DB.Update("Note", "Id", (object)entity); //<- Exception on this line
```

In the case where the custom mapper uppercase all column names, there will be a mismatch in case i.e. `entity.ID` vs the passed in primary key name `"Id"`

In PocoData,
```c#
pd.Columns = new Dictionary<string, PocoColumn>(StringComparer.OrdinalIgnoreCase);
pd.Columns.Add(primaryKeyName, new ExpandoColumn() { ColumnName = primaryKeyName });
```
we add the key as `"Id"` and also specify the `ColumnName` as `"Id"`.

So in the original for loop:
```c#
foreach (var col in (obj as IDictionary<string, object>).Keys)
{
    if (col == primaryKeyName) continue;
    pd.Columns.Add(col, new ExpandoColumn() { ColumnName = col });
}
```
`col` will never be equal to `primaryKeyName`, resulting in an `ArgumentException` with message

> An item with the same key has already been added.
since the `Dictionary` is ignoring case.

Actually I should've removed this line. It's been replaced with the key lookup:
```c#
if (col == primaryKeyName) continue;
```

So this change fixes the duplicate key bug by making sure the `ColumnName` conforms to the mapper logic.
For consistency we should technically also update the `Key`, but it's immutable and the `Dictionary` is ignoring case anyway.

Curlack:
> I know this PR is only going to an internal branch right now, not to develop, but I think the changes to `PocoData` should be their own PR -- and should stay that way going into develop. That's too significant a plumbing change to be buried in a PR about Oracle tests. Best way to handle is probably to do that on its own in a separate branch, and then once its merged to develop merge develop into this branch.

I can make the changes on develop and create a PR. How would you like to handle the changes in this PR though, first merge into develop and then pull into oracle-support?

PetaPoco/Core/PocoData.cs
asherber: Hmm, I need to think on this a little more. I get what you're saying, but something about it doesn't seem right to me. Maybe the primary key name should be run through the mapper? Maybe other fields need to be run through the mapper? Maybe it's just the caller's responsibility to pass in the correctly cased primary key name? After all, this param is not supposed to be the name of the field on the POCO; it's supposed to literally be the name of the table field.

I use an underscore mapper in my classes, so that a POCO property called `PersonID` gets mapped to a db field called `person_id`. If I'm calling that `Update()` overload, I know that I need to specify the primary key name as `'person_id`, not `PersonID`.

asherber:
The `PocoData` changes should be on their own branch with their own PR and then merged to `develop` if they're accepted. Then back out the `PocoData` changes from this branch, force push, and merge `develop` into this branch.

asherber:
Also, sorry to nitpick, but the first line of a git commit message should just be a brief (50 char) summary of what the commit does; the rest can follow after a blank line. This is because many git clients by default only display the first line of the message. See https://cbea.ms/git-commit/ for some good guidelines.

Ste1io:
I'm just sitting down to catch up on everything, been a hectic few days, but happy to see momentum is continuing at such a good clip.

PetaPoco/Core/PocoData.cs
Curlack: To be honest, I got confused most of the time about when the user should use poco casing or db casing, while working on this. So you're probably right.

My initial thought was that the dictionary keys represent the poco case (even though `PropertyInfo` is there) to help dynamic/anonymous marry strongly typed pocos, since this case won't have `PropertyInfo`, and `ColumnName` represents db case. Since that approach didn't cover all the cases, the dictionary was made to be case insensistive.

Since this part of the code is purely for dynamic, and the only affected property is the primary key, we can go both ways. Either escape the primary key (mapper should ensure original code hits exact match) or leave it to the user as long as we state it somewhere in the documentation (which I'm sure it's there somewhere)

How would you like to proceed?

Curlack: While overriding the tests I saw this line of code, which should throw if I apply the logic we discussed regarding the primary key i.e. user must provide correct casing:
```c#
DB.UpdateAsync("Note", "ID", entity); //Note: TableName is not capitalized
```

So it seems we force the user to use correct casing for primary key in order to get `PocoData`, but we escape both table name and primary key (when building the sql statement) [here](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L2506C12-L2507):
```c#
cmd.CommandText =
    $"UPDATE {_provider.EscapeTableName(tableName)} SET {sb} WHERE {_provider.EscapeSqlIdentifier(primaryKeyName)} = {_paramPrefix}{index++}";
```


> Maybe the primary key name should be run through the mapper? Maybe other fields need to be run through the mapper?

Now it looks more like we should run primary key AND other fields through the mapper.

asherber: I need to put together a couple of small concrete tests to step through this, but in principle I disagree. Mappers mediate between POCO class/property names and database table/column names. On the other hand, generally speaking, it is the caller's responsibility to make sure that anything that goes into a PetaPoco function as a string will be understood by the database. (Mappers don't peek into SQL strings, for example.) A `dynamic` object doesn't have fixed properties to map to, and so the mappers have no role.

Remember, too, that while the mapping from property name to column name is deterministic, the mapping from column name to property name is not. That is, a mapper which uppercases will always map a property called `Id` into a column called `ID`. But given a column called `ID`, there is no way to know whether the POCO property was `Id`, `id`, or `ID`. 

So I think the existing code in `PocoData.ForObject()` is correct. The incoming `obj`, as an Expando, will have received the column names as they are in the database, and in a call to something like `DB.Update()` the user is expected to pass in properly cased table and primary key names.

Curlack: Agreed. It doesn't make sense to escape `primaryKeyName` since it's not deterministic. We also assume `obj` to contain keys matching the database. There's also no guarantee that it contains the primary key.

Base on the above and the fact that the dictionary is case insensitive, I see no harm in using `TryGetValue` or `ContainsKey` to avoid adding duplicates, although the former also gives you the opportunity to fix the casing of the primary key column based on `obj`.

Are you concerned about performance?

Curlack: I've undone the PocoData changes as requested, will make them again in another branch based on develop.
Do we resolve this and continue the discussion on the coming PR?

Ste1io:
@Curlack, I'm preparing a discussion post I'll be posting here in the next few hours that directly addresses quoting behavior and identifier handling. Just wanted to let you know to avoid any possible repeated work and ensure we're on the same page in how it is handled. It's taken longer than I planned due to my absence, but also several other nuances came to the surface yesterday that factor in as well, that haven't been brought up and if not discussed now, will result in a lot of backtracking and recoding.

Curlack:
I've pushed the last changes to the remaining failing oracle tests. These tests will only pass once we have the `PocoData` changes (pending development branch PR).

Curlack:
Wanting to flex my new skills...any objections to me squashing these commits into one?

asherber:
Go for it!

Note that it's not always appropriate to squash commits. A commit should be a unit of work that has some significance. You might start out with a branch that has 6 or 7 commits, because you're following the maxim of committing often, and you might decide that really you've done two distinct things on that branch -- so you can squash down to two commits rather than one.

Curlack:
Understood. Thanks.

asherber:
Also, you shouldn't squash or otherwise change commits that have already been merged somewhere else.

One usual use case is to make local commits and squash them before pushing. Another, like this, is to have a series of commits in a pull request that get squashed before merging. GitHub can actually do that itself, from the web interface, but as the committer you have more control if you do it yourself.

PetaPoco.Tests.Integration/Databases/OracleTests/OracleUpdateTests.cs
```diff
@@ -1,14 +1,114 @@
...
+
+                CreateNote();
+                var entity = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();
+
+                entity.TEXT += " was updated";
+
+#if !NETCOREAPP
+            DB.Update("NOTE", "ID", (object)entity);
+#else
+                DB.Update("NOTE", "ID", entity);
```
Ste1io, line 67: I'd suggest doing a quick auto-format with white space diffs enabled in VS btw (may just be this file), for some of the indentation that got shifted.

PetaPoco.Tests.Integration/Databases/UpdateTests.cs
```diff
@@ -81,6 +81,8 @@ protected static void UpdateProperties(OrderLine orderLine)
             orderLine.Status = OrderLineStatus.Allocated;
         }
 
+        protected object CreateNote() => DB.Insert(_note);
+        protected Task<object> CreateNoteAsync() => DB.InsertAsync(_note);
```
Ste1io, line 85: This looked like a factory creation method initially; something more like `CreateAndInsertNote()` would be more communicative of the intent, but frankly I don't see any justification in separating it out of the test methods: it obscures what's going on in the test, and fails to serve any purpose besides forwarding the call.

Curlack: Only did it cause I didn't want to make _note protected. Similar to what you did to GetDatabase(), unless you're ok with protected. Not sure what will happend with the test data looking at the TODO statement

Ste1io: OK, makes more sense in light of that. Changing visibility to protected would be preferable at this point (and I'm doing that on other derived test classes). Overriding test methods should be tested against the same fixture as the base test methods anyways, in most cases.

Curlack: Dope! I should remove the underscore, shouldn't I?

Ste1io: That is correct. Cheers.

pleb:
@Curlack is this PR ready to merge?

Curlack:
I think so. Not sure if @Ste1io has some more changes.

Ste1io:
> @Curlack is this PR ready to merge?

I'd like to hold off till @Curlack and I can iron a couple things out with the current implementation. Been away a good bit over the Thanksgiving holiday, but will be back in the swing of things here heading into the weekend; appreciate your patience guys.

Ste1io:
@pleb @asherber @Curlack, I apologize for my extended absence over the holidays, and appreciate all of your patience.

@Curlack, I've merged the changes in the `development` branch (containing your last couple of PRs) into the public feature branch this PR is based on (`upstream feature/oracle-support`). Since this is your private feature branch, you'll want to sync it up with a rebase/force push. You can either do it locally by fetching it from upstream then force-pushing it to your origin, or you can use the option below this comment here on github, choosing the rebase option (not merge):

1703920626-CollaboratingPlatypusPetaPoco_at_featureoracle-sup

Following that, you'll need to pull the changes to your local from your origin (since your origin is what is being rebased using the update branch button.

I'll follow up this comment with another addressing this PR specifically, but wanted to get our feature branch up to date with `development` before anything else.

asherber:
Since there no conflicts between @Curlack 's branch and the base branch, I would recommend not rebasing, unless there's reason to think that recent changes in `development` will have a functional impact on the changes in this PR. From a code perspective, merging this PR will work just fine without a rebase.

cc: @Ste1io

pleb:
Apologies, I missed this one.

PR #702, opened 2023-10-19
Feature/oracle support

Curlack:
This covers all but one task. Not sure what to do for "Create a derived `OracleManagedTestProvider` class to derive from...".
These changes are for baseline Oracle setup. To confirm setup and build scripts work I ran `PetaPoco.Tests.Integration.DatabaseTests.Construct_GivenConnection_ShouldBeValid`, which executed successfully.

PR #711, opened 2023-11-03
Fix PocoData Cache Issue

### LINKED ISSUE #517: PocoData.ForType() caching

Curlack:
Include Mapper Type in cache key for PocoData. Fixes #517.

Avoid "duplicate key" side effect in `PocoData.ForObject` method when handling dynamic objects.

### DIFF
--- a/PetaPoco/Core/PocoData.cs
+++ b/PetaPoco/Core/PocoData.cs
@@ -16,7 +16,7 @@ namespace PetaPoco.Core
     {
         private static readonly object _converterLock = new object();
 
-        private static Cache<Type, PocoData> _pocoDatas = new Cache<Type, PocoData>();
+        private static Cache<Tuple<Type, Type>, PocoData> _pocoDatas = new Cache<Tuple<Type, Type>, PocoData>();
         private static List<Func<object, object>> _converters = new List<Func<object, object>>();
         private static MethodInfo fnGetValue = typeof(IDataRecord).GetMethod("GetValue", new Type[] { typeof(int) });
         private static MethodInfo fnIsDBNull = typeof(IDataRecord).GetMethod("IsDBNull");
@@ -125,7 +125,9 @@ namespace PetaPoco.Core
             if (type == typeof(System.Dynamic.ExpandoObject))
                 throw new InvalidOperationException("Cannot use dynamic types with this method");
 
-            return _pocoDatas.GetOrAdd(type, () => new PocoData(type, defaultMapper));
+            var mapper = Mappers.GetMapper(type, defaultMapper);
+            var key = new Tuple<Type, Type>(mapper.GetType(), type);
+            return _pocoDatas.GetOrAdd(key, () => new PocoData(type, defaultMapper));
         }
 
         /// <summary>
@@ -141,14 +143,17 @@ namespace PetaPoco.Core
             if (t == typeof(System.Dynamic.ExpandoObject))
             {
                 var pd = new PocoData();
+                
                 pd.TableInfo = new TableInfo();
-                pd.Columns = new Dictionary<string, PocoColumn>(StringComparer.OrdinalIgnoreCase);
-                pd.Columns.Add(primaryKeyName, new ExpandoColumn() { ColumnName = primaryKeyName });
                 pd.TableInfo.PrimaryKey = primaryKeyName;
                 pd.TableInfo.AutoIncrement = true;
+                
+                pd.Columns = new Dictionary<string, PocoColumn>(StringComparer.OrdinalIgnoreCase);
+                pd.Columns.Add(primaryKeyName, new ExpandoColumn() { ColumnName = primaryKeyName });
+                
                 foreach (var col in (obj as IDictionary<string, object>).Keys)
                 {
-                    if (col != primaryKeyName)
+                    if (!col.Equals(primaryKeyName, StringComparison.OrdinalIgnoreCase))
                         pd.Columns.Add(col, new ExpandoColumn() { ColumnName = col });
                 }
 
--- a/PetaPoco/Database.cs
+++ b/PetaPoco/Database.cs
@@ -3054,7 +3054,7 @@ namespace PetaPoco
             {
                 param.Value = DBNull.Value;
 
-                if (pocoColumn?.PropertyInfo.PropertyType.Name == "Byte[]")
+                if (pocoColumn?.PropertyInfo?.PropertyType.Name == "Byte[]")
                     param.DbType = DbType.Binary;
             }
             else


PetaPoco/Core/PocoData.cs
```diff
@@ -140,16 +141,28 @@ public static PocoData ForObject(object obj, string primaryKeyName, IMapper defa
...
-                    if (col != primaryKeyName)
-                        pd.Columns.Add(col, new ExpandoColumn() { ColumnName = col });
+                    pd.Columns.Add(col, new ExpandoColumn() { ColumnName = col });
+                }
+
+                //Ensure primary key is always in Columns list.
+                if (pd.Columns.TryGetValue(primaryKeyName, out PocoColumn column))
+                {
+                    //Prefer obj primary key name over primaryKeyName e.g. prefer obj.ID over "Id"
+                    pd.TableInfo.PrimaryKey = column.ColumnName;
```
asherber, line 159: I still don't see the need for this. The user is responsible for passing in the correct value, in the correct case, for `primaryKeyName`.

Is the issue just that the "item with the same key" message is unhelpful? Then let's make a better error message.

PetaPoco/Core/PocoData.cs
```diff
@@ -174,7 +187,7 @@ public static PocoData ForObject(object obj, string primaryKeyName, IMapper defa
         public Delegate GetFactory(string sql, string connectionString, int firstColumn, int columnCount, IDataReader reader, IMapper defaultMapper)
         {
             // Create key for cache lookup
-            var key = Tuple.Create(sql, connectionString, firstColumn, columnCount);
+            var key = Tuple.Create(defaultMapper.GetType().FullName, sql, connectionString, firstColumn, columnCount);
```
asherber, line 190: **question:** What's the use case for adding the mapper to the cache key? Under what conditions would a type being used with a given connection string potentially be used with two different mappers?

**issue:** If the mapper is needed as part of the key, we should do `var mapper = Mappers.GetMapper(Type, defaultMapper);` up here and use that mapper, not the default mapper.

PetaPoco/Core/PocoData.cs
```diff
@@ -125,7 +125,8 @@ public static PocoData ForType(Type type, IMapper defaultMapper)
             if (type == typeof(System.Dynamic.ExpandoObject))
                 throw new InvalidOperationException("Cannot use dynamic types with this method");
 
-            return _pocoDatas.GetOrAdd(type, () => new PocoData(type, defaultMapper));
+            var key = $"{defaultMapper.GetType().FullName}, {type.FullName}";
+            return _pocoDatas.GetOrAdd(key, () => new PocoData(type, defaultMapper));
```
asherber, line 129: **issue:** Need to do `var mapper = Mappers.GetMapper(Type, defaultMapper);` here and use that mapper in the key.

**question:** Are there any pros/cons to using a string key like this vs. a tuple of type and mapper type?

PetaPoco/Core/PocoData.cs
Curlack: This change is only a more robust version of the original code.
Handling incorrect casing of the primary key name is not the responsibility of this method.

PetaPoco/Core/PocoData.cs
Curlack: I just saw the same pattern as with `_pocoDatas` and thought I'd make all caches use the mapper as part of the key. Defensive programming.
I can't think of a use case where mapper would differ, so if you feel confident that this is overkill, I'll undo it.

PetaPoco/Core/PocoData.cs
Curlack: Pros: There is no need to keep the instances around (memory) and comparison is cheaper (cpu).
Cons: There could be collisions (although unlikely since we use fully qualified type names). We take a hit on readability and maintainability.

Let me use a tuple rather. There'll be consistency on the caches and the performance hit is minute.

PetaPoco/Core/PocoData.cs
```diff
@@ -124,8 +124,9 @@ public static PocoData ForType(Type type, IMapper defaultMapper)
         {
             if (type == typeof(System.Dynamic.ExpandoObject))
                 throw new InvalidOperationException("Cannot use dynamic types with this method");
-
-            return _pocoDatas.GetOrAdd(type, () => new PocoData(type, defaultMapper));
+            
+            var key = new Tuple<IMapper, Type>(Mappers.GetMapper(type, defaultMapper), type);
+            return _pocoDatas.GetOrAdd(key, () => new PocoData(type, defaultMapper));
```
asherber, line 129: I meant to use the type of the mapper, not the mapper itself (per your comment about keeping the instances around).

```csharp
var mapper = Mappers.GetMapper(type, defaultMapper);
var key = new Tuple<Type, Type>(mapper.GetType(), type);
```

PetaPoco/Core/PocoData.cs
asherber: I must be missing something. Isn't line 159 basically what you had in the other PR? Use the column name if it differs by case from `primaryKeyName`? That's what the PR description says.

Also, in what ways is this more robust?

PetaPoco/Core/PocoData.cs
asherber: Just for context, the issue with the cache for `_pocoDatas` is one that does come up occasionally, if rarely. I think you saw it while working on the integration tests. In my case, I had one POCO class that I was using against two different databases -- one which used snake_case for columns and one which used PascalCase. This meant 2 different mappers, therefore 2 different `PocoData`s.

For this change, the SQL and connection string (= database) are already part of the cache key. Even if you've got a single POCO class used for two different tables in the same database with different mapping conventions, I think the SQL itself would be different, and therefore you'd already have two different cache keys.

PetaPoco/Core/PocoData.cs
Curlack: Awe shucks! Sorry. Will fix it quick.

Curlack: I presume you want the mapper variable also? Didn't pay attention to that part until after the push.

asherber: It's just a question of readability. The single line you have is correct, but it's got a lot going on -- and most of the words are either `mapper` or `type`! Two lines makes it clearer what's going on. 

Fixing something like this is a good place for amending your commit. After you make the change, do:

```
# replaces your previous commit with a new one
git commit --amend --no-edit

git push --force
```

It doesn't matter so much, because all of your small commits will get squashed into one before merging, just a chance to try out some git tricks.

asherber: (Note that in general, while you can do whatever you want to local commits, it's not good practice to "rewrite history" by force pushing changed commits, since it can cause problems for anyone using your commits. In a case like this, though, where your PR is being reviewed and the minor change has been requested, there's no harm.)

PetaPoco/Core/PocoData.cs
Curlack: I'm going to try and elaborate on the "more robust" comment. The way things are in my head, don't always come out on paper correctly....lol.

As a whole, the method is responsible for providing a `PocoData` for the given type. For the happy path it can just return whatever comes back from the `ForType` method, but in the case of `ExpandoObejct`, the type is not useful.

The way I see it, there are two main objectives.
1. Create `Columns` from `((IDictionary<string, object>)obj).Keys`. Like a make shift `Type` if you want.
2. Ensure resulting `PocoData` instance contains the primary key in both `TableInfo.PrimaryKey` and `Columns`.

Objective 1
The closest we can get to a type is to look at the `Keys` within `obj`. which means `Keys` trump `primaryKeyName`. In this case, `primaryKeyName` is just a means to identify which `Key`, in `obj`, is the primary key, if it's there (cause there's no guarantee). So we should fill the `Columns` dictionary with `Keys` first. This avoids the complication of having to deal with duplicate key issues.

Objective 2
Since the `Columns` dictionary is not case sensitive, casing is not a factor in this method. It will be easy to figure out if we've already taken care of adding the primary key. `TryGetValue` will give us the `ExpandoColumn` if there's a key in `Columns` that matches the `primaryKeyName`. If we did not find a match, we have no choice but to use whatever the user passed as `primaryKeyName` to fill both `TableInfo.PrimaryKey` and `Columns`.

Hope this transfers my chain of thought for this change clearly and earn your approval.

PetaPoco/Core/PocoData.cs
Curlack: I've removed this change.

PetaPoco/Core/PocoData.cs
Ste1io: @asherber I'll have to circle back to it and find it, but in one of your comments you brought up some of the rationale for escaping incoming but not outgoing identifiers (column names specifically as I recall). There is actually a failing test currently (only for Oracle, as the other databases don't stress it that way), that may actually be due to (my) incorrect calling code. Will run it by you to be sure once I track down that comment and rerun the tests, but I'm fairly certain it relates to this.

asherber: I don't think that there's a good reason to privilege the keys in the Expando, because we have no idea where that Expando came from. I think you're assuming it came from `db.Fetch<dynamic>()`, but that's not necessarily true.

Since the user has specified a value for `primaryKeyName`, we should assume he knows what he's talking about and use that. If he's wrong, there will be an exception at some point, and we shouldn't try to prevent that.

I do wonder if it's reasonable to change the guard clause (line 102 in the original) from `if (col != primaryKeyName)` to `if (!col.Equals(primaryKeyName, StringComparison.OrdinalIgnoreCase))`. That takes into account the case-insensitive dictionary and get rid of the "key already exists" error message, but then the library seems to throw a null reference exception further down the line.

asherber: Oh, the NRE is just in the published library. I see that there's a fix for this in the current Oracle branch (null check on `pc.PropertyInfo`). So maybe just changing the guard clause is the better approach? Thoughts?

Curlack: Seems like it all boils down to trusting the user knows what he's doing/talking about. So the question becomes: "How is the returned PocoData instance useful to the user?"

Let's assume the user shares our sentiment that it's tedious to manually generate a dynamic object. So we can deduce that the whole object is either cased correctly or incorrectly, there is no middle ground. This means ALL keys must be treated the same way. There should be no discrimination between primary key or non-primary key. This is what my gut feeling is telling me.

I understand what you're saying about the uncertainty of where the dynamic object came from in the first place.
Let's say the object came from the user (manually generated) or some other code that auto generates it (not PP). We can safely assume that, if a key that resembles the primary key is found within the dynamic object, that the object has correctly cased keys. If no key resembles the primary key, we HAVE to assume the user knows what he's talking about and use exactly what he provided as-is.
We can also assume that the user will use this poco and the returned PocoData instance as a matching pair. 

PocoData is useful for a lot of things but, for the most part from what I could gather scanning through the code, the GetValue method lies at the core. For dynamic objects, the Columns dictionary is filled with ExpandoColumn instances, which overrides this method. If the ColumnName does not match a key in the target, it returns null.

```c#
public override object GetValue(object target)
{
    object val = null;
    (target as IDictionary<string, object>).TryGetValue(ColumnName, out val);
    return val;
}
```
Note that here the comparison is case sensitive.

Consider the following unit test:
```c#
[Theory]
[InlineData("ID", "4b8a4bb7-0104-4154-a294-6514f70f5a3e")]
[InlineData("id", "4b8a4bb7-0104-4154-a294-6514f70f5a3e")]
[InlineData("Id", "4b8a4bb7-0104-4154-a294-6514f70f5a3e")]
[InlineData("ID", null)]
[InlineData("id", null)]
[InlineData("Id", null)]
public void GetValue_GivenDynamicPoco_ShouldReturnCorrectValueWithNoSideEffects(string primaryKeyName, string primaryKeyValue)
{
    dynamic poco = new ExpandoObject();
    var dict = (IDictionary<string, object>)poco;
    var hasPrimaryKey = primaryKeyValue != null;

    //Simulate poco, consistently having ID (uppercase) primary key, irrespective of the user input, for the purposes of this test
    if (hasPrimaryKey) dict[primaryKeyName.ToUpper()] = Guid.Parse(primaryKeyValue);
    poco.CAPS = "CAPS";
    poco.TitleCase = "TitleCase";
    poco.lowercase = "lowecase";
    poco.snake_case = "snake_case";
    poco.Null = default(string);

    //Original code fails here -> ArgumentException: An item with the same key has already been added.
    PocoData pd = PocoData.ForObject(poco, primaryKeyName, new ConventionMapper());

    //This requires pd.TableInfo.PrimaryKey to be equal to pd.Columns[primaryKeyName].ColumnName
    //otherwise the primary key's key will be included in this list
    pd.UpdateColumns.ShouldBe(new[] { "CAPS", "TitleCase", "lowercase", "snake_case", "Null" });

    //This is not populated in case of dynamic objects
    pd.QueryColumns.ShouldBeNull();

    foreach (var column in pd.Columns)
    {
        var columnName = column.Value.ColumnName;

        //This will require ColumnName to be equal to poco Keys
        //if they differ GetValue will just silently return null
        object actual = column.Value.GetValue(poco);

        //We can't use dict[column.Value.ColumnName] so we'll test if it's there first
        if (dict.TryGetValue(columnName, out object expected))
            actual.ShouldBeSameAs(expected);

        if (hasPrimaryKey)
        {
            if (columnName == "Null") actual.ShouldBeNull();
            //Guard Clause fails here -> Shouldly.ShouldAssertException: actual should not be null but was
            else actual.ShouldNotBeNull();
        }
    }
}
```

Original Code
image
image
When the poco doesn't contain the primary key or if the user provides a primary key name with the correct case, the tests are happy. Otherwise we encounter the original issue, that sparked this PR, in PocoData on this line:
```c#
PocoData pd = PocoData.ForObject(poco, primaryKeyName, new ConventionMapper());
```

Guard Clause
image
image
When the poco doesn't contain the primary key or if the user provides a primary key name with the correct case, the tests are happy. Otherwise GetValue fails to return the expected value. The test fails on this line:
```c#
else actual.ShouldNotBeNull();
```

Proposed Implementation
image
Whether poco contains primary key or not, all tests pass. PocoData exactly maps the dynamic object.

There might be other usages of the PocoData that this test does not cover that might or might not highlight issues with the discrimination between primary key and other columns approach. Although I strongly believe that the proposed implementation is a way more robust approach that also compliments the usage of this library as a whole.

Man! You're really making me work for this...hahahaha. But I'm not complaining, my passion for coding is too great. I actually enjoyed this "deep" thinking exercise.

asherber: @Curlack I've read and considered your points, and I still don't agree with you. To me, it comes down to trusting what the user passes in, *especially* if it's a string. If the user says that the primary key of his table is named `id`, and he passes in (for whatever reason) a dynamic that doesn't have a property with that exact name, then he's going to have problems, and he deserves them.

I do think it's reasonable to make the guard clause in `ForObject()` case insensitive, to avoid the unhelpful "key already exists" message.

@Ste1io I think that [this line](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L3057) in the `development` branch needs an additional null propagator after `PropertyInfo`, to match the check you added in `AddParameter()`.

If both of those changes are made, then given a table with columns `ID` and `code` in a case-sensitive database (I tested in Postgres), all 3 of these tests fail with the same message, which seems correct.

```csharp
// connection string details omitted
// Postgres error: column "id" does not exist

public class MyClass
{
    public int ID { get; set; }
    public string code { get; set; }
}

[Fact]
public void Concrete()
{
    using (var db = new Database())
    {
        var c = new MyClass() { ID = 7, code = "Bar" };
        db.Update("test", "id", c);
    }
}

[Fact]
public void Anonymous()
{
    using (var db = new Database())
    {
        var a  = new { ID = 7, code = "Foo" };
        db.Update("test", "id", a);
    }
}

[Fact]
public void Dynamic()
{    
    using (var db = new Database())
    {
        var d = db.Fetch<dynamic>("select * from test").First();
        db.Update("test", "id", d);
    }
}
```

Ultimately this is @pleb's call, but I would vote for this solution (make guard clause case-insensitive, add null propagation to guard clause in `Database.SetParameterProperties()` rather than the one in this PR.

Curlack: Off screen I believe you have a mapper that lowercases table and column names?
If that's the case you could pass "Test" or even "TEST" as the table name argument and it will still be able to do the update for you.
So a user with no knowledge about what's going on under the hood, would ask: "Why is it complaining when I get the primary key name case wrong, but it doesn't care about the casing I use in the table name?"

Granted I also thought that PocoData acted like a mediator between the poco and the code interacting with the database. At least from the dynamic objects perspective. Like an "auto correct" is you want.

But let's here what @pleb thinks.

asherber: No mapper. The code above is as-is.

Curlack: Off course. Sorry. Meant to say "if you have a mapper that lowercases...".

Curlack: I'm beginning to see why you feel so string about the primaryKeyName trumping the obj.Keys.
In your tests, do you also get "Table doesn't exist" error if you case table name wrong?

asherber: A mapper isn't going to have an effect, because of [this line of code](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L2507). The `update` statement (in this case) will always use the literal primary key name passed in by the user.

I think the goal is to have dynamics work the same way as other classes; the changes I'm proposing for `dynamic` seem to accomplish that.

Yes, "table does not exist" if I use the wrong case for the table name. 

With the code in your PR, I do still get the "column does not exist" error, because I'm still passing in the wrong case in `primaryKeyName`, and that gets passed through untouched. But the PR changes other things in `PocoData.ForObject()` that I don't think need to be touched.

Curlack: Ok. One more thing. The side effect of `GetValue` silently returning null. Is that something we should be concerned about?

Curlack: > But the PR changes other things in `PocoData.ForObject()` that I don't think need to be touched.

I don't mean to be a pain, but could you please elaborate regarding the "things" you don't think need to be touched?

asherber: Seems like it's intentional for `GetValue()` to return null if it can't find the column. I don't think there have been any complaints about this, so I'd be hesitant to change established behavior.

Regarding `ForObject()`, if the guard clause is the only thing that needs to be changed in that method, then it should be the only thing that gets changed. The code from line 153 on is no longer relevant, and while I usually like object initializers, I'm not sure that nested initializers are more readable than what's there currently. If you wanted to, you could move current lines 147 and 148 up under line 144, so the `TableInfo` stuff is together.

Ste1io: This certainly appears to be one of those situations that calls for some explicit documentation for the xml doc comments, and probably a wiki page pointing out how PetaPoco handles case sensitivity when working with runtime objects. to avoid ambiguity. What exceptions are thrown, and when, might deserve some more consideration also, IMHO, as demonstrated in the two failing tests below.

While I share @asherber's concern that certain assumptions should not be made by the library, I think there is a use case where @Curlack's proposed changes would be beneficial. This may be a case of wanting to have my cake and eat it too, inside the current constraints, at least.

```cs
/*
PASS
*/
[Fact]
public virtual void Update_GivenTablePrimaryKeyNameAndDynamicType_ShouldBeValid()
{
    DB.Insert(_person);

    dynamic person = new System.Dynamic.ExpandoObject();
    person.Id = _person.Id;
    person.FullName = "Feta";
    person.Age = 19;
    person.Dob = new DateTime(1946, 1, 12, 5, 9, 4, DateTimeKind.Utc);
    person.Height = 190;

    ((int)DB.Update("People", "Id", (object)person)).ShouldBe(1);

    var personOther = DB.Single<Person>(_person.Id);
    personOther.ShouldNotBe(_person, true);
}

/*
Message:
  Shouldly.ShouldAssertException : `DB.Update("Note", "Id", (object)entity)`
    should not throw but threw
  System.ArgumentException
    with message
  "An item with the same key has already been added."

Stack Trace:
  UpdateTests.Update_GivenTablePrimaryKeyNameAndDynamicType_ShouldNotThrow() line 383
*/
[Fact]
[Trait("Issue", "#667")]
public virtual void Update_GivenTablePrimaryKeyNameAndDynamicType_ShouldNotThrow()
{
    var pd = PocoData.ForType(typeof(Note), DB.DefaultMapper);
    var tblNote = DB.Provider.EscapeTableName(pd.TableInfo.TableName);

    DB.Insert(_note);
    var entity = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();

    Should.NotThrow(() => DB.Update("Note", "Id", (object)entity));
}

/*
Message:
  Microsoft.CSharp.RuntimeBinder.RuntimeBinderException : 'System.Dynamic.ExpandoObject' does not contain a definition for 'Text'

Stack Trace:
  CallSite.Target(Closure, CallSite, Object )
  UpdateDelegates.UpdateAndExecute1[T0,TRet](CallSite site, T0 arg0)
  CallSite.Target(Closure, CallSite, Object )
  UpdateTests.Update_GivenTablePrimaryKeyNameAndDynamicType_ShouldUpdate() line 398
*/
[Fact]
public virtual void Update_GivenTablePrimaryKeyNameAndDynamicType_ShouldUpdate()
{
    var pd = PocoData.ForType(typeof(Note), DB.DefaultMapper);
    var tblNote = DB.Provider.EscapeTableName(pd.TableInfo.TableName);

    DB.Insert(_note);
    var entity = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();

    entity.Text += " was updated";

    DB.Update("Note", "Id", (object)entity);

    var entity2 = DB.Fetch<dynamic>($"SELECT * FROM {tblNote}").First();
    ((string)entity2.Text).ShouldContain("updated");
}
```

> @Ste1io I think that [this line](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L3057) in the development branch needs an additional null propagator after PropertyInfo, to match the check you added in AddParameter().

@asherber I would agree. Want me to reopen/rebase the original bugfix pr and branch with that change or just patch in a fresh pr?

Curlack: I've reduce the `ForObject()` changes to just the guard clause, move the suggested lines of code and for readability added some whitespace.

Thanks for being so patient and responding so promptly to all my questions and comments.

I really don't want to delay this PR so I can finish the Oracle integration tests, so I'm conceding, but you can't say it was because of a lack of trying hahahaha.

Would be interesting to hear @pleb's view on this  though.

Ste1io: > @Ste1io I think that [this line](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L3057) in the development branch needs an additional null propagator after PropertyInfo, to match the check you added in AddParameter().

@asherber I would agree. Want me to reopen/rebase the original bugfix pr and branch with that change or just patch in a fresh pr?

asherber: @Ste1io We can add the null propagator here; I was going to squash all of these anyway. 

I don't understand your comment above with test samples. Which commit are you running these against?

Ste1io: > I don't understand your comment above with test samples. Which commit are you running these against?

The current public feature branch for oracle support #702, which iirc has no actual core changes currently in terms of caching or mapping behavior. The tests pass for all database providers except Oracle, which exposes the key-to-column caching behavior being discussed due to the transformations to unquoted identifiers applied by Oracle. I can point out in more detail if anything's not clear, as there are numerous things at play, but didn't want to inadvertently beat a dead horse.

Curlack: I think he's referring to [this](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L2507) line. `primaryKeyName` is escaped, but the `value` provided [here](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L2455) or [here](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L2484) or [here](https://github.com/CollaboratingPlatypus/PetaPoco/blob/53e928c6dc44819aa834683abcd108029f960b77/PetaPoco/Database.cs#L2491) (depending on the scenario), is null (`GetValue()` side effect mentioned earlier).

The test is making a change to the poco and then updates the database. Then selects again from the database to confirm the update, but nothing changed cause the update affected 0 records. It affected 0 records because the statement correctly identifies "ID" as the primary key (escaping), but cannot determine the value cause "Id" was used to generate the `PocoData` instance.

Ste1io: There's nothing wrong with the assertions, the exceptions occur during the identifier lookup while accessing the fetched POCO even though the POCO was created by the library:

> I don't think that there's a good reason to privilege the keys in the Expando, because we have no idea where that Expando came from. I think you're assuming it came from db.Fetch<dynamic>(), but that's not necessarily true.
>
> Since the user has specified a value for primaryKeyName, we should assume he knows what he's talking about and use that. If he's wrong, there will be an exception at some point, and we shouldn't try to prevent that.

asherber: I would prefer not to hold this up for Oracle, since it shouldn't make anything worse for that db and may well make some things better. I suggest getting this merged and then seeing where we are.

asherber: @Curlack Since this branch has 4 tiny commits, I'm going to ask you to do an interactive rebase to squash them down to one. If you run into trouble, or just don't want to get into this, I can take care of it for you. Which git tool are you currently using? CLI, or one of the GUIs?

Curlack: I'm using Visual Studio mostly, but switch to CLI when it doesn't support some command e.g. forcing a push like I did earlier.

Curlack: Should I include the null propogator also?

asherber: @Curlack Let's continue the rebase conversation [here](https://gist.github.com/asherber/f8a452aeaf09bedca1806eb43631e9e8)

asherber (approved):

asherber:
@pleb I think this is ready for your review. Read the conversation thread only if you dare!

The gist is:

1. Add the mapper type to the `PocoData` cache key (as discussed in #517)
2. Make the guard clause for dynamics in `PocoData.ForObject()` case insensitive, to avoid needless "duplicate key" error
3. Add a null propagator to the guard clause in `Database.SetParameterProperties()`, again to avoid needless errors with dynamics

Curlack:
Thanks man

PR #714, opened 2023-12-30
add .gitattributes file for standardized line endings

### LINKED ISSUE #713: Add .gitattributes for standardized line endings
The repository currently lacks a `.gitattributes` file. This omission likely was the cause of, or at least contributed to the problems encountered in issue #503, which seem to be related to inconsistent line endings across different development environments.

Introducing a `.gitattributes` will help standardize line endings (LF/CRLF) for all contributors, regardless of their OS, and should prevent similar issues in the future.

Ste1io:
Fixes #713

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Auto detect text files and perform LF normalization
+* text=auto


asherber:
This feels a little fishy to me, something that likely isn't needed and could possibly have unintended side effects. Unless someone can show a bug which exists without this change and which is fixed by this change, I would recommend not merging it.

Ste1io:
This is standard practice, just like .gitignore and .editorconfig files. If a contributor clones a repo with Windows line endings, makes changes locally on a Linux machine, then commits the changes, you end up inconsistent line endings, which can cause subtle bugs in cases such as with the t4 templates.

https://git-scm.com/docs/gitattributes/2.5.6

asherber:
Personally, I've only ever added a `.gitattributes` file when it's been needed to solve a particular issue, not as a general part of all repos.

The default behavior of git is that all files are committed with `LF` line endings only. You'd need to take specific steps to commit a file with `CRLF`.

Ste1io:
> The default behavior of git is that all files are committed with `LF` line endings only. You'd need to take specific steps to commit a file with `CRLF`.

Assuming that the user-level git configuration does not overrides this default behavior.

To reiterate, there is no *current* issue this is fixing, as it was recently closed by #712. This is a preventative measure moving forward. I'm fairly certain that having this in place from the beginning would have prevented a lot of time and frustration for a lot of users and contributors.

If you foresee a problem with introducing this to the repo based on your experience that I haven't encountered personally, I'm definitely open to reconsideration.

Ste1io:
I've converted this PR to a draft for the time being, pending additional feedback on this. Provided it moves forward, I'll reopen after ensuring the files are correctly normalized.

asherber:
I don't feel that strongly about it; I'm just a little wary about changing a configuration option without a good reason. And I'm not convinced that the lack of a `.gitattributes` has caused any issues for this repo.

As far as I can tell, the text files in this repo are all committed with `LF` line endings, as expected.

pleb:
This has always caused a little pain, but usually just with diff'ing when the diff logic doesn't ignore line endings. But this isn't a project concern but a user concern.

With the T4 templates, however, I think from memory the line ends must be `crlf` (windows) or it doesn't work. T4 is super picky. 

I'm happy to add a .gitattributes file.

Maybe we can follow the [github docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) for this 

```
# Set the default behaviour in case people don't have core.autocrlf set.
* text=auto

# The above will handle all files NOT found below

# These files are text and should be normalized (Convert lf => crlf)
*.tt text eol=crlf
*.ttinclude text eol=crlf
```

asherber:
@pleb Your comment about the `tt` files needing `CRLF` is what I glean from #713. However, as best as I can tell, the file currently in the repo just have `LF` (download a zip of the repo from GH and check those files), and the latest comments in the issue indicate that the problem does not currently exist. That's why I wasn't sure this was needed right now.

pleb:
@asherber agree. Let's leave this for now, and if it becomes an issue in the future, we address it again. 

The T4 templates are probably due to being retired anyway, as I'm unsure of their value over @hippasus's https://github.com/hippasus/PetaPoco.DBEntityGenerator. Plus, given we have a replacement and the fact they're super annoying to use and maintain I'd vote for their removal

PR #720, opened 2024-05-11
Update Npgsql to v4.1.13

asherber:
Npgsql has a high-severity vulnerability, which they take seriously enough to issue new builds of out-of-support versions. It doesn't matter that much for this project, since the dependency is only in the integration tests project, but merging this will make Dependabot happy.

As a side note, it would probably be a good idea to go through the test projects and update frameworks and libraries to more recent versions. I can take a look at this if I can get the Docker stuff running on my new machine.


pleb:
@asherber I had a little trouble running the solution on my Mac, and I don't readily have access to a Windows machine. 

I've tested the docker-compose file, and it works for me

